PR #17, opened 2014-02-11
Combination of all my issue ideas.

dubiousdavid:
Separated out vector and list.  Use vector for 'as' and list for function calls.  Default behavior for select, join, where, group, order-by, having, etc. is to append rather than replace.  Merge if you like :)


seancorfield:
At this point, given the history and direction of the library, and that this pull request has conflicts (and can't be merged), I'm inclined to reject it -- and ask that individual issues be opened for discussion if you still want to pursue them?

PR #31, opened 2014-11-05
Add full join

justindell:


ddellacosta:
@jkk @MichaelBlume Seems reasonable to support this as its part of the SQL ANSI standard.  Furthermore this patch seems nice and clean and has tests too--would propose merging it in if you folks don't object.

src/honeysql/helpers.clj
```diff
@@ -93,6 +93,12 @@
 (defhelper merge-right-join [m clauses]
   (update-in m [:right-join] concat clauses))
 
+(defhelper full-join [m clauses]
+  (assoc m :full-join clauses))
+
+(defhelper merge-full-join [m clauses]
+  (assoc m :merge-full-join clauses))
```
MichaelBlume: As discussed on https://github.com/MichaelBlume/honeysql/pull/2, this needs fixing

test/honeysql/core_test.clj
```diff
@@ -34,6 +35,7 @@
             :join [:draq [:= :f.b :draq.x]]
             :left-join [[:clod :c] [:= :f.a :c.d]]
             :right-join [:bock [:= :bock.z :c.e]]
+            :full-join [:beck [:= :beck.x :c.y]]
```
MichaelBlume: Would be good to see a merge-full-join test as well

MichaelBlume:
merge-full-join needs fixing, otherwise I'm ready to merge

PR #34, opened 2014-12-30
Support basic common table expressions.

akhudek:
Note that this requires modifying the clause-order, thus users cannot use the built in extensibility alone to support CTEs.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -156,7 +156,7 @@
 
 (def clause-order
   "Determines the order that clauses will be placed within generated SQL"
-  [:with :select :insert-into :update :delete-from :columns :set :from :join
+  [:with :with-recursive :select :insert-into :update :delete-from :columns :set :from :join
    :left-join :right-join :where :group-by :having :order-by :limit :offset
    :values :query-values])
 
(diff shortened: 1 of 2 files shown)


MichaelBlume:
So what I find interesting about this pull is that it almost -- almost! -- doesn't look like it needs to go in honeysql to be useful. Almost all of the code is defmethods, or helpers for defmethods. You could put those in your own project, or in a standalone "honeysql extension" library and require them before you start constructing any SQL. The one hitch is clause-order -- it's just a constant. I wonder if it would make sense to set up some mechanism for extending clause-order, maybe by registering new clauses with a 'priority' that could be sorted on, as Haskell does for order-of-operations. Then changes like this one could be substantially more self-serve.

jkk:
Great idea @MichaelBlume. I'll create an issue for making clause order extensible in some way

ddellacosta:
@jkk @MichaelBlume I think pull request #43 is awesome and is a no-brainer.  However I also don't see why HoneySQL shouldn't support CTEs--they've been a part of the ANSI SQL standard for some time and are widely supported by many databases (http://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL).  What if we first merged in #43 then refactored this pull request to suit?

MichaelBlume:
Yeah, makes sense to me

ddellacosta:
Updated to work with the changes introduced in #43, and added a few very basic tests.

PR #35, opened 2015-01-07
support postgres-native parameters

icambron:
It would be nice to use alternative ways of creating parameters; specifically, I have in mind Postgres, which uses `$1`, `$2`, etc instead of `?`, `?`, etc. (See [here](http://www.postgresql.org/docs/9.4/static/xfunc-sql.html)). This is useful when using HoneySQL in conjunction with a non-JDBC driver, such as [this one](https://github.com/alaisi/postgres.async). Anyway, this patch allows the user to specify that they want Postgres-native parameters:

```cljs
(sql/format :params {:foo "bar"} :parameterizer :postgresql)
```

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -30,6 +30,8 @@
 
 (def ^:dynamic *param-counter* nil)
 
+(def ^:dynamic *all-param-counter* nil)
+
 (def ^:dynamic *input-params* nil)
 
 (def ^:dynamic *fn-context?* false)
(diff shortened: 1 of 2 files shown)


MichaelBlume:
@ddellacosta @jkk what do you think of this? On the one hand, it's postgres only, but on the other, this would be pretty hard to package as an extension.

MichaelBlume:
Anyhow, if anyone's interested, I've attempted to fix the merge conflicts here https://github.com/MichaelBlume/honeysql/commit/params-merge

icambron:
Hey - @MichaelBlume thanks for taking on the conflicts. If more is needed from me to get this merged (or, I suppose, rejected), please let me know.

PR #36, opened 2015-01-09
Support for `RETURNING` keyword

erikprice:
PostgreSQL supports this clause on INSERT statements so as to return things such as generated IDs, etc.

### DIFF
--- a/src/honeysql/core.clj
+++ b/src/honeysql/core.clj
@@ -49,7 +49,8 @@
       :query-values
       :update
       :set
-      :delete-from"
+      :delete-from
+      :returning"
   [& clauses]
   (let [[base clauses] (if (map? (first clauses))
                          [(first clauses) (rest clauses)]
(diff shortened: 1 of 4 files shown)


erikprice:
Updated to remove redundant `defmethod` per comment from @MichaelBlume .

ddellacosta:
I'm on the fence about this one, but I think it would be good to try to keep HoneySQL limited to ANSI SQL at this point--allowing in vendor-specific syntax could be a slippery slope, and ideally I think HoneySQL should simply make it easy to extend it when one has to support a vendor-specific bit of syntax.  Related: would #42 allow for this one to be provided as an extension?

MichaelBlume:
Yeah, that's pretty much exactly what I was thinking, I'd rather keep
vendor-specific syntax out of the core lib. And yes, with #43 merged this
basically becomes an extension.

One thing I could *maybe* see doing is having, say, a honeysql.postgresql
ns that could be required separately and include postgres-specific syntax,
but I'd also be happy simply limiting the library to ANSI SQL.

MichaelBlume:
Ha, didn't see your edit, but we seem to be on the same wavelength @ddellacosta =)

ddellacosta:
Yes!

MichaelBlume:
Yeah, I'm going to go ahead and close this one for now. Note again that with v0.5.0 this will be very easy to add from your own project.

PR #37, opened 2015-01-23
memoize sqlable?

MichaelBlume:
In simple benchmarks, calls to sql/format spend the majority of their time in satisfies. Make the simple cases simpler with caching.


MichaelBlume:
A call to format in my app went from 350 microseconds to 50 microseconds with this change (using criterium)

I can put up a test project with benchmarks if desired.

MichaelBlume:
Actually, I've got a better solution to this

PR #38, opened 2015-01-23
extend ToSql to cover Object

MichaelBlume:
This commit avoids extraneous calls to satisfies?. Satisfies? is very
slow compared to protocol method dispatch, because method dispatch is
cached and satisfies? is not. Instead of using satisfies? to check for
cases where we need to fall back to a default behavior, we extend ToSql
to java.lang.Object, providing a default behavior directly.

This commit boosts honeysql's speed substantially. In my benchmarks,
80-90% of the time spent calling sql/format was spent in satisfies?.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -258,26 +258,26 @@
                    (paren-wrap sql-str)
                    sql-str)))
   nil
-  (-to-sql [x] "NULL"))
+  (-to-sql [x] "NULL")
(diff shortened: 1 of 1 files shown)


MichaelBlume:
Times for the giant map in the unit tests

with change: 83µs
without change (master): 405µs

MichaelBlume:
Rewrote this as a less invasive patch.

PR #40, opened 2015-02-20
Added support for casts of the form CAST(foo AS type)

senior:


ddellacosta:
@jkk @MichaelBlume Standard ANSI SQL cast--don't see any reason not to merge this in.

PR #41, opened 2015-02-26
avoid name collision with clojure.core/update

MichaelBlume:
present in 1.7, currently in alpha


MichaelBlume:
Maybe premature since the problem is only in alpha

PR #45, opened 2015-03-03
Update README.md

cloojure:
Put in automatic lein coordinates from clojars


MichaelBlume:
I like these because they auto-maintain themselves, I slightly dislike them because they prevent the user from copy-pasting from the github front page.

ddellacosta:
Ha, I have the exact same feeling...but in the end I think it's a win to have stuff updated automatically, so I'd be for merging it in.

PR #48, opened 2015-03-04
Simplify type implementations using defrecord

MichaelBlume:
This removes a decent amout of boilerplate from honeysql.types by using defrecord. In particular, it removes the need for custom printing methods.

I'll be honest, my motivation for working on this change is a bit esoteric -- in my vim sessions I have C-t bound to clojure.tools.namespace.repl/refresh and clojure.test/run-all-tests. The trouble is that refresh recreates the types in honeysql.types and this winds up breaking the round-trip printing test (in particular, the problem is the custom readers -- they wind up bound to the un-reloaded versions of the reader functions which are bound to the old types). Using default print/read for records fixes this for me.

I've left in the old read methods for backwards compatibility.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -245,6 +245,16 @@
                    (apply fn-handler fn-name (.args x)))))
   SqlRaw
   (-to-sql [x] (.s x))
+  SqlParam
+  (-to-sql [x] (let [pname (param-name x)
+                     x (if (map? @*input-params*)
+                         (get @*input-params* pname)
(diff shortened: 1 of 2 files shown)


ddellacosta:
Sorry, still trying to wrap my head around everything that's going on here...will respond in more detail later today/tomorrow.

MichaelBlume:
It may help to look at the three commits separately

MichaelBlume:
Closing this for now, some of it might be worth revisiting, but #52 solves the real problem.

PR #49, opened 2015-03-04
in unit test, put expected value first

MichaelBlume:
is assumes that the first argument to = is a fixture, and the second is
the result of a computation, when it prints Expected: and Actual:


ddellacosta:
+1

Issue #33, opened 2014-12-01
more INSERT ... SELECT


mccraigmccraig:
INSERT INTO where one of the inserted columns is a subquery is supported with :

    (-> (insert-into :user_profile_to_role)
          (values [{:user_profile_id user-id
                    :role_id         (-> (select :id)
                                         (from :role)
                                         (where [:= :name role-name]))}])

and INSERT INTO where all of the inserted columns match a subquery with (a happy accident ?):

    (-> (insert-into [:bar (-> (select :id) (from :role))]))

but i can't figure out a way to to an INSERT...SELECT where the columns to be inserted are specified, e.g.

  INSERT INTO foo (id) (SELECT id FROM bar);

is this currently supported, or do i need to wrap the sub-query to match columns with the INSERT INTO table ?

ddellacosta:
This is resolved with #50, now merged in and released.

PR #50, opened 2015-03-04
Supporting extended INSERT INTO...SELECT syntax

ddellacosta:
I propose supporting insert queries in the form of

```sql
INSERT INTO foo (a, b, c) SELECT d, e, f FROM baz
```

...using the format:

```clojure
{:insert-into [[:foo [:a :b :c]] {:select [:d :e :f] :from [:baz]}]}
```

So, for example you could do:

```clojure
(hsql/format
  (hsql/build :insert-into [[:foo [:a :b :c]]
                            (hsql/build :select [:d :e :f] :from :baz)]))

;; ["INSERT INTO foo (a, b, c) (SELECT d, e, f FROM baz)"]
```

This is an extension to the insert-into method already available.  This would address #33.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -46,6 +46,7 @@
 
 (defn- undasherize [s]
   (string/replace s "-" "_"))
+
 (defn quote-identifier [x & {:keys [style split] :or {split true}}]
   (let [qf (if style
              (quote-fns style)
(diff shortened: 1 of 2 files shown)


MichaelBlume:
Looks good to me

ddellacosta:
@jkk Going to merge this in unless you object--it simply extends what is already there to cover another case (which is still ANSI SQL), so I think it falls within the scope of the lib adequately.

ddellacosta:
\o/

PR #55, opened 2015-03-15
support postgres-native parameters

MichaelBlume:
duplication of #35 with merge conflicts fixed and a couple small simplifications.


MichaelBlume:
Basically I like this, even though it's only for postgresql, it's not that invasive, and it helps pg users in a way that can't be added through extension. I'd be in favor of merging. @jkk, @ddellacosta?

ddellacosta:
Yep, I agree with the thinking here!

PR #59, opened 2015-03-16
use StringBuilder, not concatenation

MichaelBlume:
Don't merge this just yet, it's definitely an experiment.

Wanted to see what would happen if we tried to use a single StringBuilder for formatting sql instead of repeatedly concatenating strings. What happens is the code gets pretty gnarly, honestly. I'd say we *probably* shouldn't merge this unless we have actual speed complaints from users. Query formatting does get faster (85 microseconds -> 75 microseconds) on the map in the unit tests. I imagine we'd see more substantial gains on queries with lots of nested function calls.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -8,15 +8,6 @@
 
 ;;;;
 
-(defn comma-join [s]
-  (string/join ", " s))
-
-(defn space-join [s]
-  (string/join " " s))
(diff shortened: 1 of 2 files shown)


MichaelBlume:
Another reason not to merge this is that it will break third-party extensions that return strings instead of appending to ```*builder*```

PR #60, opened 2015-03-16
Added DELETE clause that takes a FROM clause

MichaelBlume:
duplicate of #27, less the merge conflicts


MichaelBlume:
Still needs tests, of course

MichaelBlume:
@ddellacosta @Menthalion , as I said on #61, if either of you feel like adding tests, we can get this in sooner.

seancorfield:
*bump* Tests? @ddellacosta @Menthalion ?

seancorfield:
I pulled this down locally but I can't see how it is supposed to work. `delete` seems to only take a single table (since it's identical to `delete-from` except it omits `"FROM"` from the generated SQL).

I modified it locally to have a plural argument `tables` and `(map to-sql tables)` and that seems to produce the expected results. Will put up a new PR (*with tests!*) shortly.

PR #64, opened 2015-03-26
Adds :enable-values-subqueries? option to honeysql.format/format

emidln:
When :enable-values-subqueries? is set to false, treat maps provided
in a :values clause as a parameter instead of as a subquery. By default,
:enable-values-subqueries? is true, retaining existing behavior.

This allows handling IPersistentMaps at the jdbc level. Vectors could 
probably also use this treatment, but that can be for another pull request.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -38,6 +38,11 @@
 
 (def ^:dynamic *subquery?* false)
 
+(def ^:dynamic *handling-values* false)
+(def ^:dynamic *enable-values-subqueries?* 
+  "When False, subqueries are not computed for IPersistentMap values"
+  true)
+
 (def ^:private quote-fns
   {:ansi #(str \" % \")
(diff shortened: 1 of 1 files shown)


MichaelBlume:
Yep, looks good to me. @jkk @ddellacosta ?

MichaelBlume:
As an alternative what if we did something like https://github.com/MichaelBlume/honeysql/commit/value and expected the user to wrap the map value in a call to value?

emidln:
I'd be all for fewer dynamic bindings. It was a bit rough wading into this file...

emidln:
As a nice bonus, value would let us handle misc other types if we wanted (e.g. Vector to turn into an org.postgresql.jdbc.Jdbc4Array).

emidln:
Withdrawing this in favor of https://github.com/jkk/honeysql/pull/65.

Issue #63, opened 2015-03-26
Opt of sub-queries for IPersistentMap in ToSql 


emidln:
Right now, ToSql defines a method (-to-sql [x] ...) that is specified to IPersistentMap in such a way as to prevent passing JSON values with (honeysql.helpers/values [{:my-json-field {:foo "bar"}}]). This is a significant obstacle for me right now. Would a patch that adds an optional second parameter (meaningful to at least the IPersistentMap implementation) be considered?

MichaelBlume:
Yep, that sounds like an annoying problem, I'd be interested in seeing a patch to fix it.

MichaelBlume:
Actually I wonder if it makes sense for IPersistentMap to implement ToSql at all. Is the IPM implementation used recursively or only as an entry point? If it's only an entry point would it make sense to have it be its own function?

emidln:
It's unfortunately used recursively to process all of the clauses.

emidln:
See https://github.com/jkk/honeysql/pull/64 for a pull request implementing this as an option to honeysql.format/format

MichaelBlume:
Done, and released in 0.5.2 =)

PR #65, opened 2015-03-27
Provide a wrapper to prevent maps/vectors from being read as subqueries

MichaelBlume:
alternative to #64 attempts to solve #63 

Is value the best name? ```raw-value```? ```not-a-query-no-really```? something else?

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -247,6 +247,17 @@
         (into [sql-str] @*params*)
         [sql-str]))))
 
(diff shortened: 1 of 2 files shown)


emidln:
After having played with this for a few days, I'd prefer this. It's extremely unusual for me to want to disable subqueries in preference to values in general, and I can simply wrap the JSON/Arrays I need with values.

As to the bikeshedding, I don't mind `value`. It's namespaced already, which seems good enough to me.

PR #68, opened 2015-04-15
Sql array

loganmhb:
I had occasion to use the ANSI SQL array syntax yesterday and HoneySQL didn't support it, so I had to make some raw-sql calls and couldn't use bind parameters. I decided to try out adding it to the library and these are the results, in case this seems like something valuable enough to be incorporated.


MichaelBlume:
Nice =)

PR #69, opened 2015-04-17
NULL order specification support

mishok13:
SQL:2003 standard has introduced optional ability to specify the order in which NULL values are handled in ORDER BY clause. This pull request implements said functionality.

The only databases that support this functionality currently are PostgreSQL (since version 8.3) and Oracle (since at least 10g).

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ Here's a big, complicated query. Note that Honey SQL makes no attempt to verify 
              [:between :f.e 10 20]]
     :group-by [:f.a]
     :having [:< 0 :f.e]
-    :order-by [[:b.baz :desc] :c.quux]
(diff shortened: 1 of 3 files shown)


MichaelBlume:
This looks good to me. Is the syntax definitely the same in Postgres and Oracle?

mishok13:
@MichaelBlume unfortunately I don't have Oracle to test this out, but as far as the documentation goes, it seems to be the same both in PostgreSQL and Oracle: http://www.postgresql.org/docs/9.4/static/queries-order.html and http://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqlj13658.html

MichaelBlume:
Cool, thanks =)

PR #70, opened 2015-04-17
Support locking selects

dball:
I use honeysql for a few projects and would like to use it to generate locking selects. My primary database is mysql, alas, but I've generalized the lock map value to support the primary modes for mysql and postgresql. Extension for novel lock clauses is possible via a multimethod.

If you're interested in merging this, I'll be happy to add documentation, specific tests for all the to-sql cases, and the helper.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -188,6 +188,7 @@
    :order-by 190
    :limit 200
    :offset 210
+   :lock 215
    :values 220
    :query-values 230})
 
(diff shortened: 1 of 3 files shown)


MichaelBlume:
Hm, ok, is this in the ANSI spec at all, or is this vendor-specific for mysql and postgres?

dball:
As far as I can tell, FOR UPDATE is part of the ANSI SQL standard, though I confess I can't actually find a copy of the standard on the web and I'm not about to pay the ANSI committee for a copy. Nonetheless, it seems to be supported in MySQL, PostgreSQL, MS SQL, and Oracle.

Alas, MySQL and PostgreSQL have different syntaxes for requesting shared locks, and PostgreSQL is the only one with the special NOWAIT hint, which is why I opted to describe the lock request in an open map instead of a simple keyword or a variant tuple or the like, and to be explicit about the vendor share mode in the ones I built in for MySQL and PostgreSQL.

MichaelBlume:
Ok, I think I'm sold then =)

Issue #73, opened 2015-04-20
The IN infix operator does not work with named query parameters


dball:
This form doesn't work as expected:

```
(def hsql {:select [:*] :from [:customers] :where [:in :id :?ids]})
(honeysql.core/format hsql {:ids [1 2 3]}) ; => ["SELECT * FROM customers WHERE (id in ?) " [1 2 3]]
```

If this were to work, it would have to result in something of the form:

```
["SELECT * FROM customers WHERE (id IN (?,?,?))" 1 2 3]]
```

If there's any interest in making this work, I'd be happy to take a stab at it. It's unclear offhand if the current behavior is desirable for some reason though, so I thought I'd ask first.

MichaelBlume:
Thanks for the report!

Yeah, I'd be interested in seeing this work, please feel free to take a stab at it =)

ddellacosta:
I think some of this simply has to do with the fact that there are so many ways to do it--for example, here's a workaround for your example:

```clojure
;; using: (def hsql {:select [:*] :from [:customers] :where [:in :id :?ids]})

user=> (honeysql.core/format (assoc-in hsql [:where 2] [1, 2, 3]))
["SELECT * FROM customers WHERE (id in (1, 2, 3))"]
user=> (honeysql.core/format (assoc-in hsql [:where 2] ["one", "two", "three"]))
["SELECT * FROM customers WHERE (id in (?, ?, ?))" "one" "two" "three"]
```

But yeah, seems like something that worked correctly using your example code would be nice.

dball:
Yeah, I just use inline values for the most part without issue, but I am in the process of experimenting with extracting some hsql queries into vars and the param version would be the most readable form. I'll see if I can come up with something.

MichaelBlume:
Fixed by #74

PR #74, opened 2015-04-21
Apply seq to sets when converting to sql

dball:
This allows them to be used as values, e.g. for in clauses, as demonstrated in the test. The second commit goes further and provides a fix for #73, albeit with a question on the correct way to handle *param-names*.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## 0.5.3 In development
 
+* Convert seq param values to literal sql param lists (@dball)
 * Apply seq to sets when converting to sql (@dball)
 * Support locking selects (@dball)
 * Add sql array type and reader literal (@loganmhb)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 ## 0.5.3 In development
 
+* Apply seq to sets when converting to sql (@dball)
 * Support locking selects (@dball)
 * Add sql array type and reader literal (@loganmhb)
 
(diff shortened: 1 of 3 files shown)


MichaelBlume:
This looks entirely reasonable to me, but I'd kind of like to get another +1. @ddellacosta? @jkk?

ddellacosta:
The first commit seems straightforward and good, and I like the Parameterizable protocol, and the fact that there are tests.  I'm also wondering if the Parameterizable protocol could help simplify the code for checking opts vs. params in the beginning of format's defn code...but I digress.

Anyways, no objections here, looks good to me!

dball:
What of the question about *param-names*? This commit doesn't do anything special with it when expanding *params* with a literal param array, but I can't divine the intent of the *param-names* collection anyway. I see @jkk added it in c8fbc16ec3815888597a788c13906aaca1032022 but that's all I've been able to figure. Perhaps it's purely a debugging aid?

MichaelBlume:
Hey, sorry for the silence there.

The thing is I don't have a good handle on what *param-names* is doing, so I'm a little leery of throwing *param-names* and *params* out of sync as this seems to do. @jkk could you perhaps give us a little guidance on what use case return-param-names was meant to serve?

MichaelBlume:
(don't worry about the conflict, I'll just merge this by hand and push when we're ready to merge)

dball:
Absent clarification from @jkk, I think the sensible thing to do would be to repeat the repeated param name once for each value in the IN sequence.

MichaelBlume:
So something like https://github.com/MichaelBlume/honeysql/commit/dupe-pnames ?

Issue #75, opened 2015-04-24
Support for exists keyword


cddr:
e.g.

```
exists ( select 1 from foo )
```

or

```
select *
   from foo
 where exists ( select 1
                           from bar where foo.id = bar.foo_id )
```

MichaelBlume:
fixed in #76

PR #76, opened 2015-04-24
[#75] Add support for exists

cddr:
Hi @jkk,

This PR implements #75.

The exists clause is useful for efficiently testing for existence without forcing the database to realize all the rows and then count them. If all we need to know is whether some relation exists, it's enough to find the first item in the relation and return true immediately.

IMO, it also makes some queries that would otherwise be joins more intuitive. For example to find "currently logged in users", you might use something like this....

```
select id
  from users
 where exists ( select 1 from sessions where user.id = session.user_id )
```
instead of
```
select id
  from users
inner join sessions on user.id = session.user_id
```
Please let me know if you'd like any changes or additional test coverage.

Thanks,
Andy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ pom.xml*
 .lein-plugins
 .classpath
 .project
+.nrepl-port
 bin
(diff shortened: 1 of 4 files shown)


MichaelBlume:
Looks good to me.

PR #81, opened 2015-06-25
Add '@@' as an infix operator for postgres text search

adreyer:
'@@' is an infix operator to test whether a query matches a search document.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -73,7 +73,7 @@
 (def infix-fns
   #{"+" "-" "*" "/" "%" "mod" "|" "&" "^"
     "and" "or" "xor"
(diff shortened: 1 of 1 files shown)


ddellacosta:
I would argue this should just be added as an extension in your local project vs. adding this to HoneySQL, at least until we decide to add a PostgreSQL-specific namespace or something (which will provide justification for adding SQL Server, Oracle, MySQL, etc. namespaces which seems probably like a path we don't want to go down, in my opinion).

We do this in our project via the following:

```clojure
;; tsvector_matches
(defmethod fn-handler "@@"
  [_ a b & more]
  (if (seq more)
    (hsql.format/expand-binary-ops "@@" a b more)
    (str (to-sql a) " @@ " (to-sql b))))
```

PR #82, opened 2015-07-22
quoting of AS clauses should not be splitted on "."

DTB01:
quoting of AS clauses should not be splitted on "."

Running

``` clojure
(-> (select [:ad.id "myad.id"]) (from :ad) (sql/format :quoting :mysql) )
```

produces

``` sql
["SELECT `ad`.`id` AS `myad`.`id` FROM `ad`"]
```

Which causes a MySQLSyntaxErrorException. The correct quoting would be

``` sql
["SELECT `ad`.`id` AS `myad.id` FROM `ad`"]
```

Enclosing the whole alias like this, works for both ansi and oracle style quoting(ansi accepts both splitting and not splitting)

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -303,7 +303,7 @@
              " AS "
              " ")
            (if (string? (second x))
-             (quote-identifier (second x))
+             (quote-identifier (second x) :split false)
              (to-sql (second x))))))
   SqlCall
   (to-sql [x]
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -85,7 +85,9 @@
    "regex" "regexp"})
 
 (defprotocol ToSql
-  (to-sql [x]))
+  (to-sql
+    [x]
+    [x & args]))
 
 (defmulti fn-handler (fn [op & args] op))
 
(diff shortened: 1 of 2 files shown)


DTB01:
Hi guys
Anyone looking at this ? It's a bugfix(for MySql SQL), and a pretty straightforward checkin...

Cheers!

MichaelBlume:
Hey, sorry for the delay, I was a bit bewildered by the use of varargs in a protocol, which I wouldn't actually have expected to work.

MichaelBlume:
Ok, this works kind of by coincidence. defprotocol reads

```clj
(defprotocol ToSql
  (to-sql
    [x]
    [x & args]))
```

Not as "to-sql has one or more args" but as "to-sql has one arg (named x) or three args (named x, &, and args)". Now, it happens that when you call to-sql with varargs, you are using three args, so it basically works out. Still, this is going to be broken in an upcoming Clojure if this patch lands: http://dev.clojure.org/jira/browse/CLJ-888). I'd like to find a way to organize this code that doesn't involve faking a varargs protocol.

seancorfield:
@DTB01 Can you provide an updated/correct PR for this given @MichaelBlume 's comments? Otherwise, this is going to be closed out (even tho' the underlying issue is still correct).

seancorfield:
Closing out, since over a month has passed. The underlying bug #216 will be fixed in 0.9.4.

PR #83, opened 2015-08-06
Add fn-handler for CASE statement

loganlinn:
Support `CASE` statements with `call`


ddellacosta:
This conforms to SQL-92 and seems like a no-brainer to merge in-- @jkk @MichaelBlume or anyone else object?

dball:
:+1: I could use this

PR #87, opened 2015-08-23
Fix the union clause

dball:
The previous version omits the UNION between the base query and the
first of the union queries

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## 0.6.2 In development
 
+* Fix union clause (@dball)
+
 ## 0.6.1
 
 * Define parameterizable protocol on nil (@dball)
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -519,7 +519,7 @@
   (str "WITH RECURSIVE " (comma-join (map cte->sql ctes))))
 
 (defmethod format-clause :union [[_ maps] _]
-  (string/join " UNION " (map to-sql maps)))
+  (str "UNION " (string/join " UNION " (map to-sql maps))))
 
 (defmethod format-clause :union-all [[_ maps] _]
-  (string/join " UNION ALL " (map to-sql maps)))
+  (str "UNION ALL " (string/join " UNION ALL " (map to-sql maps))))


ddellacosta:
I'm quite confused by this.  As far as I can tell the `UNION` and `UNION ALL` functionality is working fine:

```clojure
user> (require '[honeysql.core :as hsql] '[honeysql.format :as fmt] :reload)
nil
user> (hsql/format {:union [{:select [:foo] :from [:bar1]}
                            {:select [:foo] :from [:bar2]}]})
["(SELECT foo FROM bar1) UNION (SELECT foo FROM bar2)"]
user> (hsql/format {:union-all [{:select [:foo] :from [:bar1]}
                                {:select [:foo] :from [:bar2]}]})
["(SELECT foo FROM bar1) UNION ALL (SELECT foo FROM bar2)"]
user>
```

Am I missing something?

dball:
The problem is there's no UNION or UNION ALL operator between the basis and the unions:

```
syringe.repl=> (honeysql.core/format {:select [:id] :from [:foo] :union [{:select [:id] :from [:bar]}]})
["SELECT id FROM foo (SELECT id FROM bar)"]
```

ddellacosta:
Semantically a [union operation](https://en.wikipedia.org/wiki/Set_operations_%28SQL%29#UNION_operator) doesn't have any sort of "basis," multiple sets simply get merged into a single set.  This is fundamental to set theory, reflected by how it works in SQL, and the way it is represented in honeysql now maps cleanly to that.

So you can achieve what you need to with the syntax that is already supported by the `:union` keyword as it exists now.  But most problematically, what you have added is going to break current functionality for anyone currently using this.

dball:
https://dev.mysql.com/doc/refman/5.5/en/union.html

We have to patch the union clause in production because what honeysql is producing is not valid SQL.

dball:
Oh, what you're suggesting is that the outermost query map has a single :union keyword with as seq of selects as the value; that makes sense.

ddellacosta:
Yeah, exactly, the vector of `:selects` is an analog to the set of result sets that is getting merged by the UNION operation.

This should be better documented and tested though.  I've [added some tests](https://github.com/jkk/honeysql/pull/88) at least, I'll work on the documentation as soon as I get some time...

PR #89, opened 2015-08-25
Document the correct use of the :union clause

dball:
My co-workers and I naively expected to include the :union in the
primary query, not as a container. This is a good representation, but
merits explanation.


ddellacosta:
Thanks!

PR #92, opened 2015-10-11
Added support for preserving dashes in quoted names.

jrdoane:
This pull request is to resolve the following issue by adding optional functionality.
https://github.com/jkk/honeysql/issues/80

This is what normally happens:
```clojure
(format {:select [:*] :from [[:some-new-schema.user-accounts :ua]] :where [:= :ua.user-id 12345]} :quoting :ansi)
;;; Yields the following:
["SELECT * FROM \"some_new_schema\".\"user_accounts\" \"ua\" WHERE \"ua\".\"user_id\" = 12345"]
```

This change supports doing the following while allowing nothing to change if the ```:allow-dashed-names``` option is not provided.

```clojure
(format {:select [:*] :from [[:some-new-schema.user-accounts :ua]] :where [:= :ua.user-id 12345]} :quoting :ansi :allow-dashed-names true)
;;; Results in:
["SELECT * FROM \"some-new-schema\".\"user-accounts\" \"ua\" WHERE \"ua\".\"user-id\" = 12345"]
```

I think that this change will make a lot of PostgreSQL developers happy. I know it will for me. :)

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -38,6 +38,8 @@
 
 (def ^:dynamic *subquery?* false)
 
+(def ^:dynamic *allow-dashed-names?* true)
+
 (def ^:private quote-fns
   {:ansi #(str \" % \")
    :mysql #(str \` % \`)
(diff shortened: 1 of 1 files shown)


jrdoane:
Hmmm. I'll see about updating the failing test that checks specifically if de-dashing works.

jrdoane:
Found the issue. Default was backwards. I'll do a new pull request with the change.

PR #93, opened 2015-10-11
Added support for preserving dashes in quoted names.

jrdoane:
*** Take 2 ***
This pull request is to resolve the following issue by adding optional functionality.
https://github.com/jkk/honeysql/issues/80

This is what normally happens:
```clojure
(format {:select [:*] :from [[:some-new-schema.user-accounts :ua]] :where [:= :ua.user-id 12345]} :quoting :ansi)
;;; Yields the following:
["SELECT * FROM \"some_new_schema\".\"user_accounts\" \"ua\" WHERE \"ua\".\"user_id\" = 12345"]
```

This change supports doing the following while allowing nothing to change if the ```:allow-dashed-names?``` option is not provided.

```clojure
(format {:select [:*] :from [[:some-new-schema.user-accounts :ua]] :where [:= :ua.user-id 12345]} :quoting :ansi :allow-dashed-names? true)
;;; Results in:
["SELECT * FROM \"some-new-schema\".\"user-accounts\" \"ua\" WHERE \"ua\".\"user-id\" = 12345"]
```

I think that this change will make a lot of PostgreSQL developers happy. I know it will for me. :)

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -232,7 +232,7 @@
               *input-params* (atom params)
               *quote-identifier-fn* (quote-fns (:quoting opts))
               *parameterizer* (parameterizers (or (:parameterizer opts) :jdbc))
-              *allow-dashed-names?* (:allow-dashed-names opts)]
+              *allow-dashed-names?* (:allow-dashed-names? opts)]
       (let [sql-str (to-sql sql-map)]
         (if (seq @*params*)
           (if (:return-param-names opts)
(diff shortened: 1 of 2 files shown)


jrdoane:
@MichaelBlume : Let me know if there is anything you don't like.

MichaelBlume:
Yep, looks good to me. @jkk @ddellacosta any objections? If not I'll merge/release.

ddellacosta:
@MichaelBlume sorry for the slow response--looks good to me, seems like a useful and important thing to have in here!  I would say go ahead and merge it in.

PR #95, opened 2015-10-12
Support column names in :with clauses

### LINKED ISSUE #84: Support for column names in WITH
At least in PostgreSQL, when creating CTEs with `WITH`, column names can be passed in (to produce e.g. `WITH m (col1, col2) AS VALUES (2,3), (3,4) SELECT col1 FROM m` -- or more usefully to pass in multiple columns of data for an `UPDATE`, in my experience. If I'm understanding honeysql's code correctly this isn't supported, and it'd be useful for it to be!

emidln:
This binds `honeysql.format/*subquery?*` to `false` which turns
on processing of a cte-name like `[:foo {:columns [:a :b :c]}]`
resulting in a string like `WITH foo (a, b, c)`.

```clj
{:with [[[:static {:columns [:a :b :c]}]
         {:values [[1 2 3] [4 5 6]]}]]}
```

```sql
WITH static (a, b, c) AS (VALUES (1, 2, 3), (4, 5, 6))
```

This would fix #84

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -507,10 +507,13 @@
 
 (defmethod format-clause :delete-from [[_ table] _]
   (str "DELETE FROM " (to-sql table)))
-  
+
 (defn cte->sql
   [[cte-name query]]
-  (str (to-sql cte-name) " AS " (to-sql query)))
+  (str (binding [*subquery?* false]
+         (to-sql cte-name))
+       " AS "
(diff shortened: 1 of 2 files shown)


MichaelBlume:
Looks ok to me, @jkk , @ddellacosta ?

ddellacosta:
@MichaelBlume Looks good to me.

Issue #100, opened 2015-11-09
Support for `returning` clause for insert & update


niquola:

ddellacosta:
Closing this since the corresponding PR is closed.

PR #101, opened 2015-11-09
add returning clause

niquola:
related to #100


niquola:
Ups, i see it's duplicate

Issue #102, opened 2015-11-25
Is there a way to generate intersect queries ? 


bcambel:
My assumption was that they would be very similar to union and union-all queries but I assume they're not implemented yet. Is that the case ? 

Thanks, Bahadir

jakemcc:
@bcambel Just pulled out our extension to honeysql that adds intersect support and submitted a PR.

Pretty straight forward to extend in your own project though.

```clojure
;; honeysql.format required as `hformat` in files `ns` form.

(defmethod hformat/format-clause :intersect
  [[_ maps] _]
  (string/join " INTERSECT " (map hformat/to-sql maps)))
```

bcambel:
ah great, thanks @jakemcc!

PR #103, opened 2015-11-29
add intersect support

jakemcc:
PR adds support of adding `:intersect` clauses to queries. I had already extended honeysql to add support for :intersect and saw that others were looking for it.

Should solve #102.


ddellacosta:
@jkk @MichaelBlume gonna go ahead and merge this in unless you object in the next few days.  Seems consistent with having UNION and UNION-ALL.

PR #105, opened 2016-01-03
Made Honey SQL helpers compatible with ClojureScript

azapps:
I am using the Honey SQL helpers in ClojureScript to construct the Query map which is used later on the backend to query the database (I know that this is in most cases not a good idea).

This is the change I made so if you like it, feel free to merge it to master.

Downside: This breaks compatibility with Clojure < 1.7


azapps:
This is obsolete due to #130

PR #108, opened 2016-02-09
prevent records from being treated as maps

### LINKED ISSUE #106: Single, unaliased sql/raw seems to be broken in 0.6.2
I looked through the commits between 0.6.1 and 0.6.2 and it wasn't obvious to me what would have broken it, but here's the behavior I'm seeing:

```clojure
(-> (select (sql/raw "1 + 1"))
    (from :foo)
    (sql/format))

;; 0.6.1
;; => ["SELECT 1 + 1 FROM foo"]

;; 0.6.2
;; => ["1 + 1"]
```

Adding another column or an alias fixes the problem:

```clojure

(-> (select :a (sql/raw "1 + 1"))
    (from :foo)
    (sql/format))

;; => ["SELECT a, 1 + 1 FROM foo"]

(-> (select [(sql/raw "1 + 1") :a])
    (from :foo)
    (sql/format))

;; => ["SELECT 1 + 1 AS a FROM foo"]
```

MichaelBlume:
Thanks to @dwhitt and @blinsay for reporting and for providing
test cases.

Fixes #106

### DIFF
--- a/src/honeysql/helpers.clj
+++ b/src/honeysql/helpers.clj
@@ -7,12 +7,17 @@
 (defmethod build-clause :default [_ m & args]
   m)
 
+(defn plain-map? [m]
+  (and
+    (map? m)
+    (not (instance? clojure.lang.IRecord m))))
+
 (defmacro defhelper [helper arglist & more]
   (let [kw (keyword (name helper))]
     `(do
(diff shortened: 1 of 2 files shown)


blinsay:
thanks!

djwhitt:
Excellent, thanks for tracking that down!

Issue #109, opened 2016-02-24
quoted identifiers should escape their quote character


csummers:
Quoted identifiers should allow the quoted character within identifier output.  This requires that the appropriate quote character be escaped per the type of `:quoting`.

So, this is what is currently happening:

``` clj
user=> (-> (sql/build :select (keyword "column_with_\"_double_quote")) (sql/format :quoting :ansi))
["SELECT \"column_with_\"_double_quote\""]
```

and this is what should happen:

``` clj
user=> (-> (sql/build :select (keyword "column_with_\"_double_quote")) (sql/format :quoting :ansi))
["SELECT \"column_with_\"\"_double_quote\""]
```

`:ansi` should replace `"` with `""`
`:mysql` should replace `backtick` with `double backtick` *
`:mssql` should replace `]` with `]]`

 \* _after a few tries I couldn't get the markdown to render a double back tick right together_

csummers:
Created pull request #110 for this.

In the initial report I should have said `:sqlserver` instead of `:mssql`.  I also accounted for the `:oracle` quoting in the pull request.

This issue is potentially an SQL injection attack vector if identifiers are taken from user input and expect `quote-identifier` to do the right thing.

Thanks!

csummers:
Looks like this rolled with 0.7.0, so closing.  Thanks!

PR #110, opened 2016-02-24
escape quote characters in quoted identifiers

csummers:
Pull request for issue #109

`quote-identifier` now escapes the quote character.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -41,10 +41,10 @@
 (def ^:dynamic *allow-dashed-names?* false)
 
 (def ^:private quote-fns
-  {:ansi #(str \" % \")
-   :mysql #(str \` % \`)
-   :sqlserver #(str \[ % \])
-   :oracle #(str \" % \")})
+  {:ansi #(str \" (string/replace % "\"" "\"\"") \")
+   :mysql #(str \` (string/replace % "`" "``") \`)
+   :sqlserver #(str \[ (string/replace % "]" "]]") \])
+   :oracle #(str \" (string/replace % "\"" "\"\"") \")})
 
 (def ^:private parameterizers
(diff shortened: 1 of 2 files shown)


ddellacosta:
@jkk going to merge this in shortly unless you object.  I don't love that there's vendor specific stuff in here but it's relatively isolated and was already in the codebase, so...best to make it more secure if so.

csummers:
I know the documentation is quite scarce for escaping quoted identifiers, so here are a few references.

It's in the SQL spec.  See the BNF starting at [delimited-identifier](http://savage.net.au/SQL/sql-99.bnf.html#delimited identifier):

```
<delimited identifier>    ::=   <double quote> <delimited identifier body> <double quote>
<delimited identifier body>    ::=   <delimited identifier part> ...
<delimited identifier part>    ::=   <nondoublequote character> | <doublequote symbol>
<nondoublequote character>    ::=   !! (See the Syntax Rules)
Note that the two successive double quote characters must have no other character (such as a space) between them. The lexical analyzer would normally deal with this sort of issue.
```

Here's a page referencing MS SQL Server [Delimited Identifiers](https://msdn.microsoft.com/en-us/library/ms176027.aspx):

> ...you can create a table with the name Employee], where the closing square bracket is part of the name. To do this you have to escape the closing square bracket using two more square brackets...

And Mysql [Identifiers](http://dev.mysql.com/doc/refman/5.7/en/identifiers.html):

> Identifier quote characters can be included within an identifier if you quote the identifier. If the character to be included within the identifier is the same as that used to quote the identifier itself, then you need to double the character.


P.S. I'm the author of [HugSQL](https://github.com/layerware/hugsql), so I just recently dealt with these issues, and I have given quoting in HugSQL the same behavior as in this pull request.

PR #111, opened 2016-03-01
allow joins without a pred

stuarth:
simple change to `format-join` supporting `nil` predicates

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -414,10 +414,10 @@
   (str "WHERE " (format-predicate* pred)))
 
 (defn format-join [type table pred]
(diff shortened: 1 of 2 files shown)


MichaelBlume:
Maybe make the predicate argument actually optional instead of requiring the user to pass nil?

stuarth:
@MichaelBlume not sure we can make the predicate optional without making the join syntax ambiguous, i.e. in `(join :a [:foo :f])` how do we determine whether `[:foo :f]` is a table or predicate?

Any ideas? Am I misreading?

MichaelBlume:
Aha, makes sense.

Issue #112, opened 2016-03-22
Select distinct on is broken


piranha:
Issue #1 was closed because you can do `{:select [(sql/call :distinct-on :x) :y]}`, but that generates SQL which is not working in Postgres: `select distinct on (x), y`. Postgres complains that the comma is invalid syntax.

I wonder if it's still possible to do `DISTINCT ON` somehow.

lambdahands:
I'm running into this issue as well. I thought I might be able to solve this by using the `:modifiers` key, but the current [behavior](https://github.com/jkk/honeysql/blob/36a40bc67cf021267f7a3cb01f57987bf193c93b/src/honeysql/format.cljc#L465) simply uppercases the keys within the value, which is embedded in the `format-clause` implementation of `:select`. I chose to create a new implementation `:selectd` with added behavior to support `DISTINCT ON (...)`:

```clj
(require '[honeysql.format :as f])
(require '[clojure.string :as string])

(defmethod f/format-clause :selectd [[_ fields] sql-map]
  (str "SELECT "
       (when (:distinct-on sql-map)
         (str "DISTINCT ON ("
              (f/comma-join (map name (:distinct-on sql-map)))
              ") "))
       (f/comma-join (map f/to-sql fields))))
```

It's used like so:

```clj
{:selectd     [:foo]
 :from        [:bar]
 :distinct-on [:baz]
```

This workaround has its limits as we lose the composability of HoneySQL with `:selectd` (the helper methods like `merge-select` no longer work here).

I think there are two ways to solve this:

1. The `:select` clause could be augmented to handle a `:distinct-on` top-level clause, which I believe would violate the general design decision of the library to be database agnostic (`DISTINCT ON` is specific to Postgres).
2. Open up the `:modifiers` to multi-methods, and allow users to extend the space between `SELECT` and its clauses as needed.

The second option seems intuitive to me. Something like this might work:

```clj
{:select    [:foo]
 :from      [:bar]
 :modifiers [[:distinct-on :baz]]}
```

piranha:
I think so, it's just unclear if it's possible to make it work, because in our codebase we're doing this: `{:select [(sql/raw "distinct on (id) id") :other_field]}`.

seancorfield:
Given HoneySQL's position of trying to be vendor-neutral and the presence of reasonable approaches to make this work (without changing HoneySQL), I'm closing this out.

PR #113, opened 2016-04-05
Fix distinct on

krish8664:
This is a fix for Issue #112 - where the `DISTINCT ON` wasn't being generated properly.

Please do let me know, if something has to be changed or if it needs more tests.

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -17,6 +17,14 @@
 (defn paren-wrap [x]
   (str "(" x ")"))
 
+;; Checks if the first is a `DISTINCT ON` statement, if yes then return a string
+;; with an appended space at the end of the first and then comma joining the rest
+;; else simply return a comma joined string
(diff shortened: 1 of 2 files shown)


ddellacosta:
Sorry I haven't commented on this until now.

So in keeping with the general philosophy of HoneySQL, this seems a bit too PostgreSQL-specific to me.  Even in the PostgreSQL docs you will find:

> The DISTINCT ON clause is not part of the SQL standard and is sometimes considered bad style because of the potentially indeterminate nature of its results. With judicious use of GROUP BY and subqueries in FROM, this construct can be avoided, but it is often the most convenient alternative.

Think this could be added as an extension somehow to https://github.com/nilenso/honeysql-postgres ?

seancorfield:
This should be part of `honeysql-postgres`, not the core library.

krish8664:
Thanks. Closing this.

PR #114, opened 2016-04-07
Add postgres extension - upsert & returning

krish8664:
Added `UPSERT` and `RETURNING` clause to a PostgresSQL specific namespace. There were a few suggestions that vendor specific clauses should be in a separate namespace so decided to add this here.  

The `upsert` would be 
```clojure
(sql/format 
  {:insert-into :distributors
   :values [{:did 10 :dname "Pinp Design"}
                {:did 11 :dname "Foo Bar Works"}]
   :upsert {:on-conflict :did
                :do-update-set [:dname]}})
=> ["INSERT INTO distributors (did, dname) VALUES (10, ?), (11, ?) ON CONFLICT (did) DO UPDATE SET dname = EXCLUDED.dname" "Pinp Design" "Foo Bar Works"]
```

also `returning`, which would be
```clojure
(sql/format 
  {:delete-from :distributors
   :where [:> :did :10]
   :returning [:*] })
=> ["DELETE FROM distributors WHERE did > 10 RETURNING *"]
```
which would be the same as #96 .

If this approach is acceptable, then I could add it for a few other vendors as well, if not then I am willing to spend sometime change it provided a better implementation.

### DIFF
--- a/src/honeysql/postgres/format.clj
+++ b/src/honeysql/postgres/format.clj
@@ -1,0 +1,61 @@
+(ns honeysql.postgres.format
+  (:refer-clojure :exclude [format])
+  (:require [honeysql.format :refer :all]))
+
+;; Move the whole default priorities here ?
+(def postgres-clause-priorities
+  {:upsert 225
+   :on-conflict 230
+   :on-conflict-constraint 230
+   :do-update-set 235
+   :do-update-set! 235
+   :do-nothing 235
+   :returning 240
+   :query-values 250})
+
+;; FIXME : Not sure if this is the best way to implement this, but since the `clause-store` is being used
+;; by format to decide the order of clauses, not really sure what would be a better implementation.
+(doseq [[k v] postgres-clause-priorities]
+  (register-clause! k v))
+
+(defn get-first [x]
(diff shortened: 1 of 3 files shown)


ddellacosta:
I'm definitely not opposed to this but based on previous conversations think the consensus is that this (a postgres-specific set of honeysql helpers) should be a separate lib.

@MichaelBlume or @jkk or anyone else?

krish8664:
Okay, makes sense. 

 I guess I can move this to a separate library to include postgres specific helpers.

krish8664:
So I finally created [honeysql-postgres](https://github.com/nilenso/honeysql-postgres) with some of the features I needed, I have also added basic `create-table` and `drop-table` which might not be exactly postgres specific.

I plan on adding more postgres specific feature/functions to it down the line. In the mean time, if anyone has any pointers or suggestions do let me know :)

ddellacosta:
@krish8664 Finally following up on this--I've got a PR up to add a link to your lib in the main README.  If I don't hear from @MichaelBlume or @jkk or anyone else, I'll merge it in a few days.  Otherwise going to close this one.  Thanks!

Issue #115, opened 2016-04-14
Unable to quote function calls


jrdoane:
In PostgreSQL, custom functions may be quoted and there are certain characters what require quoting in PostgreSQL (when there are hyphens, upper case letters, symbols, or spaces.) So it may not be unrealistic for me to want generate SQL that looks like this:
```
SELECT "non-default-schema"."upsert-email-address"('jdoane@example.io');
```

This is valid PostgreSQL SQL however, the quoting is required because of the existence of special characters (in this case, namely the hyphen.) If I wrote this using HoneySQL I would have done something like this:
```clj
(do
  (require 'honeysql.core)
  (def q {:select [(sql/call :non-default-schema.upsert-email-address "jdoane@example.io")]})
  (honeysql.core/format q :quoting :ansi :allow-dashed-names true))

=> ["SELECT non-default-schema.upsert-email-address(?)" "jdoane@example.io"]
```

As you can see, neither the schema or the function call itself got escaped. I also don't have any mechanism to enable quoting when a ```SqlCall``` is used.

I'm not sure if this is PostgreSQL specific functionality but, having control over what exactly gets quoted might be a nice feature to have because there are cases where quoting is not required and others where it is and I'm fairly certain this isn't unique to PostgreSQL (although, behavior between databases might differ.)

The most sensible solution to me seems to create another record for ```SqlCallQuoted``` and extending the ```ToSql``` protocol to support dispatch on the new ```SqlCallQuoted``` record that will do what I'm expecting.

...am I on the right track here, am I over thinking the problem, and is this a bug or expected (and valid,) behavior for HoneySQL given the provided code above?

seancorfield:
Enhancing the quoting feels reasonable to me but this definitely needs more analysis (the PR helps).

PR #116, opened 2016-04-14
F/quoted functions

jrdoane:
See #115.

This adds support for being able to quote database function calls. This was required for PostgreSQL because user defined functions can be fully qualified and must be quoted if the name contains certain characters. As a result, I updated HoneySQL to support the following which I will quote from the addition I made to the README:

In some cases, you may want to quote function calls. To do this, you can use the ```sql/call-quoted``` function instead of the ```sql/call``` function.

```clj
(sql/format
  {:select (sql/quoted-call :my-schema.some-function 1 2 3)}
  :quoting :ansi :allow-quoted-names? true)
=> ["SELECT \"my-schema\".\"some-function\"(1, 2, 3)"]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,16 @@ To be able to use dashes in quoted names, you can pass ```:allow-dashed-names tr
 => ["SELECT \"f\".\"foo-id\", \"f\".\"foo-name\" FROM \"foo-bar\" \"f\" WHERE \"f\".\"foo-id\" = 12345"]
 ```
 
+In some cases, you may want to quote function calls. To do this, you can use the
+```sql/call-quoted``` function instead of the ```sql/call``` function.
+
+```clj
+(sql/format
+  {:select (sql/quoted-call :my-schema.some-function 1 2 3)}
+  :quoting :ansi :allow-quoted-names? true)
+=> ["SELECT \"my-schema\".\"some-function\"(1, 2, 3)"]
+```
+
 Here's a big, complicated query. Note that Honey SQL makes no attempt to verify that your queries make any sense. It merely renders surface syntax.
 
 ```clj
--- a/src/honeysql/core.clj
+++ b/src/honeysql/core.clj
@@ -7,6 +7,7 @@
             [clojure.string :as string]))
 
 (defalias call types/call)
+(defalias call-quoted types/call-quoted)
 (defalias raw types/raw)
 (defalias param types/param)
 (defalias format format/format)
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -312,9 +312,9 @@
   SqlCall
   (to-sql [x]
     (binding [*fn-context?* true]
-       (let [fn-name (name (.-name x))
-             fn-name (fn-aliases fn-name fn-name)]
-         (apply fn-handler fn-name (.-args x)))))
+       (let [quoted? (.-quoted? x)
+             fn-name ((if quoted? to-sql name) (.-name x))]
+         (apply fn-handler (fn-aliases fn-name fn-name) (.-args x)))))
   SqlRaw
   (to-sql [x] (.-s x))
   clojure.lang.IPersistentMap
--- a/src/honeysql/types.clj
+++ b/src/honeysql/types.clj
@@ -1,18 +1,24 @@
 (ns honeysql.types)
 
-(defrecord SqlCall [name args])
+(defrecord SqlCall [name args quoted?])
 
 (defn call
   "Represents a SQL function call. Name should be a keyword."
   [name & args]
-  (SqlCall. name args))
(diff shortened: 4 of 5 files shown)


jrdoane:
Is there any reason why this hasn't been considered yet?

ddellacosta:
@jrdoane Apologies. I'm going to spend some time reviewing outstanding stuff in HoneySQL this weekend--will get back to you very soon I hope, if someone else doesn't first!

ddellacosta:
@jrdoane So, I finally had a chance to poke at this a bit and think about it.  I have a few concerns/questions:

- why not simply use:

```clojure
honeysql.format> (hsql/format (hsql/build :select [(hsql/raw "\"non-default-schema\".\"upsert-email-address\"('jdoane@example.io')")]))
["SELECT \"non-default-schema\".\"upsert-email-address\"('jdoane@example.io')"]
honeysql.format>
```
That is, are there situations where this is necessary where `raw` wouldn't suffice?

Secondarily, and probably more importantly, it seems like quoted identifiers if they are added as a feature should extend to more than simply function calls.  It seems to me that it would be less ad-hoc and probably a better longer-term approach to figure out how to simply add quoting to *all* identifiers, not just function calls.  That way I could do something like (using the PostgreSQL doc's example from https://www.postgresql.org/docs/9.5/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS):

```clojure
honeysql.format> (hsql/format (hsql/build :update :$my_table set {:$a 5}))
["UPDATE \"my_table\" SET \"a\" = ?" 5]
honeysql.format>
```

Extending from there, you could then ideally have this kind of syntax for the example in your original issue:
```clojure
honeysql.format> (hsql/format (hsql/build :select (hsql/call :$non-default-schema.upsert-email-address "jdoane@example.io")))
["SELECT \"non-default-schema\".\"upsert-email-address\"(?)" "jdoane@example.io"]
honeysql.format>
```

I'm using `$` as one idea of how we could extend keywords in the way we already do to achieve this.

Considering that this seems to be broadly applicable across SQL variants, and as far as I can tell is supported by the ANSI standard, that seems like a better long-term approach.  Thoughts?

jrdoane:
@ddellacosta That's interesting. I had a long car ride so I was thinking about this for a little bit. If we were to push this a little further, what about ```.```? It's not invalid for a column name to include not only periods but whitespace as well, at least in PostgreSQL (not that I would ever condone doing this in practicality.) It does make me wonder if something a little more general should be considered. As it stands right now, if you have a keyword like ```:this.is.a.column```, HoneySQL will split by every period and quote everything else. It would seem to me that being able to explicitly state "this is my schema" and "this is my name" might give a dev more control over this. I really don't know what that should look like but, it's a consideration as well. A solution to that problem might be to only split on the first or last occurrence of a period but, that feels kind of hackish.

With that all said, I still like your idea about doing something like ```:$non-default-schema.upsert-email-address``` but, how about going a little further. We could use two symbols to denote "this is the schema, this is the name" or, we could give the dev control over what token is used to split the identifier apart instead of merely splitting on a period.

How about something like this:
```clojure
(hsql/call :$.non-default-schema.upsert-email-address)
```
The $ would be used to say this identifier is indended to be quoted, the second character says "this is the  token I want to split by." So if someone for some unknown reason needs a period in a field name, they can have control over what character it used. On the same token, that might be overcomplicating it but, I definitely could see using a period in a column name if someone is going to be using something like clojure.spec in upcoming Clojure 1.9 where namespaced keywords are pretty important.

Either way, I do like your idea and it probably makes more sense.

seancorfield:
@jrdoane @ddellacosta This definitely needs revisiting against the latest version of HoneySQL because it does, now, support quoting on nearly all identifiers, but the function call is still excluded:
```clojure
user=> (h/format (select (sql/call :this-is-a.test-fn)) :quoting :ansi :allow-dashed-names? true)
["SELECT this-is-a.test-fn()"] ; function is not quoted
user=> (h/format (-> (select (sql/call :this-is-a.test-fn)) (from :table)) :quoting :ansi :allow-dashed-names? true)
["SELECT this-is-a.test-fn() FROM \"table\""] ; table is quoted
user=> (h/format (h/build :update :table :set {:a 5})) ; update example above
["UPDATE table SET a = ?" 5]
user=> (h/format (h/build :update :table :set {:a 5}) :quoting :ansi)
["UPDATE \"table\" SET \"a\" = ?" 5] ; both table and a are quoted
```
So it seems like functions are still considered a special case, exempt from quoting but I suspect we do not wanted them quoted in the general case (I haven't tried whether it would be legal in MySQL/SQL Server)?

Is `#sql/raw` sufficient for the (few?) cases where a function call _would_ need quoting?

PR #122, opened 2016-05-17
Treat numbers as paramterized values to fix issues with NaN, Infinity…

akhudek:
…, and -Infinity.

See https://github.com/jkk/honeysql/issues/121

### DIFF
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -290,8 +290,6 @@
         (quote-identifier x))))
   clojure.lang.Symbol
   (to-sql [x] (quote-identifier x))
-  java.lang.Number
-  (to-sql [x] (str x))
   java.lang.Boolean
   (to-sql [x]
     (if x "TRUE" "FALSE"))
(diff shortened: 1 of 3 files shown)


akhudek:
Yeck, let me look into the errors.

akhudek:
Alright, this broke a ton of tests since they all assume numbers are put directly into the html. Before I take all the time to update the tests, can someone comment on if this change would be acceptable?

MichaelBlume:
This looks reasonable to me. Is there any reason numbers were specially
handled in the first place, other than reducing the number of parameters?

akhudek:
The only two things I can think of would be performance and for things like limit, offset or "select 1 from foo;" but so far all of those seem to work fine. I've tested this patch with our quite large code base and so far I haven't run into any problems.

MichaelBlume:
I'm gonna snooze this another 24 hours in case the other maintainers have concerns, then merge.

ddellacosta:
I've thought about this a bit FWIW and can't really come up with any compelling arguments against.  I suppose it's possible that performance could be negatively impacted at scale if you have a ton of numeric params where you didn't before but feel like it's probably a negligible concern unless we hear otherwise from library users (and I happen to be familiar enough with Alex's codebase that I suspect that if he doesn't see any issues, we're probably good).

So, voting to get this merged in.

Issue #131, opened 2016-07-09
Using fields with forward slashes chops off namespace.


jrdoane:
Since we're living in the Clojure world, it might not be unrealistic that we would want fields in the database to look and feel like idiomatic names we might see in Clojure.

Lets say I had a database where I had a table called `system.tables` that contains keys that look like `schema/oid`, `table/oid`,  `tablespace/oid`, `user/oid`, `user/name`, `schema/name`, `table/name` and, `tablespace/name`.

This all works fine for a query like this:

``` clojure
{:select [:*] :from [:system.tables]}
```

...but fails when doing something like this:

``` clojure
{:select [:table/oid :schema/oid :user/oid] :from [:system.tables] :where [:= :tablespace/name "foo"]}
```

In every case, the namespace on the keyword is discarded and strictly the name portion of the keyword remains. So in the example above with ansi quoting and allow-dashed-names? enabled, I would see something like this:

``` clojure
["SELECT \"oid\", \"oid\", \"oid\" FROM \"system\".\"tables\" WHERE \"name\" = ?" "foo"]
```

Is there any particular reason why this is the case and would an update to HoneySQL to fix this be welcomed? It's an edge case until you decide to do it, then it will probably show up everywhere.

seancorfield:
PR #132 was merged so closing this.

PR #132, opened 2016-07-09
Allow namespaced keywords and symbols

jrdoane:
This pull request solves this problem: https://github.com/jkk/honeysql/issues/131

This should preserve the namespace portion of a keyword if it is used in a query. This will allow developers to use field names that contain forward slashes by using namespaced keywords and symbols in Clojure.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## 0.7.1 In development
 
+* Allow namespaced keywords and symbols for queries. (@jrdoane)
+
 ## 0.7.0
 
 * Parameterize numbers, properly handle NaN, Infinity, -Infinity (@akhudek)
--- a/src/honeysql/format.clj
+++ b/src/honeysql/format.clj
@@ -62,7 +62,10 @@
              (quote-fns style)
              *quote-identifier-fn*)
         s (cond
-            (or (keyword? x) (symbol? x)) (name-transform-fn (name x))
+            (or (keyword? x) (symbol? x))
+            (name-transform-fn
+              (str (when-let [n (namespace x)]
+                     (str n "/")) (name x)))
             (string? x) (if qf x (name-transform-fn x))
             :else (str x))]
     (if-not qf
--- a/test/honeysql/format_test.clj
+++ b/test/honeysql/format_test.clj
@@ -29,6 +29,10 @@
     (is (= (quote-identifier :foo-bar.moo-bar :style :ansi)
            "\"foo-bar\".\"moo-bar\""))))
 
+(deftest test-namespaced-identifier
+  (is (= (quote-identifier :foo/bar) "foo/bar"))
+  (is (= (quote-identifier :foo/bar :style :ansi) "\"foo/bar\"")))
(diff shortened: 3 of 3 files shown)


ddellacosta:
I can't really find any reason to object to this-- @MichaelBlume unless you or someone else objects will merge it in.

jeaye:
Any update on this?

MichaelBlume:
Looks good to me, conflict will need to be fixed

jeaye:
I think this wouldn't work with nested namespaces though, since the `foo.bar` syntax is typically reserved for looking up columns in tables. I see that the tests included in this PR don't handle that case either.

I think that's something, since this is merged, that should be tested and/or, if it's not supported, documented.

MichaelBlume:
@jeaye could you provide a sample test? I'm not sure I know the use case you're pointing at.

jeaye:
@MichaelBlume To illustrate what I'm raising: what would be the semantic meaning of `:my-program.util.md5/foo` for SQL?

To answer my own question, we can look at this: http://stackoverflow.com/a/21419181 Now, my point is that mapping nested namespaces to SQL dot notation simply doesn't make sense to me and I'm concerned whether or not it should be supported by honeysql.

MichaelBlume:
I might be misunderstanding you, but that sounds like a GIGO problem? If the keyword is namespaced to something that makes sense on the left side of the slash, it'll work, if it doesn't, it won't, no?

jeaye:
I think you get me. My stance, really, is that overloading qualified keywords to have SQL's dot notation semantics is abusive and worse than just using strings -- strings don't have an existing meaning, when containing something like `my-db.my-schema.my-table`; keywords do. Furthermore, the usage of `:anything.core/foo` presupposes the existence of that namespace, at some point.

It irks me; doesn't bother you?

MichaelBlume:
Not super much? We're in a DSL and the keywords basically mean what we say they mean. @ddellacosta  @jkk any thoughts?

MichaelBlume:
Note that this PR is currently in 0.9.0-beta1 but not in any release version

bredtape:
Looking forward to this in the next release ;).

MichaelBlume:
It has now been in a few releases, unless I'm missing something...

bredtape:
My mistake. I thought it was not released because I used keywords with namespaces, but that was due to the namespace containing .

Example (with Postgres as the intended target), no quoting:
`(sql/format {:insert-into :inventory :values [{:a.b/f 3 :e/g 5}]})`
> ["INSERT INTO inventory (a.b/f, e/g) VALUES (?, ?)" 3 5]

... but each namespace element is quoted with :ansi
`(sql/format {:insert-into :inventory :values [{:a.b/f 3 :e/g 5}]} :quoting :ansi)`
> ["INSERT INTO \\"inventory\\" (\\"a\\".\\"b/f\\", \\"e/g\\") VALUES (?, ?) 3 5]

seancorfield:
Given this is merged, it seems right to close #131?

Issue #138, opened 2016-08-24
Unable to order by in UNION query


madvas:
Let's say we have honeysql query like this

``` clojure
(honeysql.format/format
    {:union-all [{:select [:amount :id :created_on]
                  :from [:transactions]}
                 {:select [:amount :id :created_on]
                  :from [:other_transactions]
                  :order-by [[:created_on :desc]]}]})
```

It compiles into 

``` clojure
["(SELECT amount, id, created_on FROM transactions) 
UNION ALL 
(SELECT amount, id, created_on FROM other_transactions ORDER BY created_on DESC)"]
```

But actually in order to ORDER BY working correctly by combining both queries it has to be out of parenthesis. Like: 

``` clojure
["(SELECT amount, id, created_on FROM transactions) 
UNION ALL 
(SELECT amount, id, created_on FROM other_transactions) 
ORDER BY created_on DESC"]
```

How to achieve this in honeysql?

seancorfield:
You'd need:
```clj
{:union-all [{:select ...}
             {:select ...}]
 :order-by [...]}
```
(because you want the `ORDER BY` on the whole query, not just the second subquery).

However, there seems to be a bug with `UNION` and `ORDER BY` that I just ran into where this generates the `ORDER BY` (and any `LIMIT` etc) at the start of the formatted query, rather than at the end.

seancorfield:
PR #140 submitted to address the bug I highlighted above.

rnewman:
See also #142, which removes the non-standard parentheses from `UNION` and `INTERSECT`, which I suspect will also fix this problem!

For the benefit of folks ending up here through searches — the correct standard way of doing `ORDER BY` in the various parts of a `UNION` is through subqueries:

```
SELECT amount, id, created_on FROM transactions
UNION ALL
SELECT amount, id, created_on FROM
  (SELECT amount, id, created_on FROM other_transactions ORDER BY amount DESC LIMIT 5)
ORDER BY amount ASC
```

which in honeysql is:

```
             {:union
              [{:select [:amount :id :created_on]
                :from [:transactions]}
               {:select [:amount :id :created_on]
                :from [{:select [:amount :id :created_on]
                        :from [:other_transactions]
                        :order-by [[:amount :desc]]
                        :limit 5}]}]
              :order-by [[:amount :asc]]}
```

There's a test for that in #142.

seancorfield:
[...]

PR #140, opened 2016-09-14
Address #138 : add priority for union/union-all

seancorfield:
By prioritizing union / union-all to happen early on, complex queries
with order by / limit / offset etc will be correctly formatted _after_
the union / union-all clause has been formatted.

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -174,7 +174,9 @@
 
 (def default-clause-priorities
   "Determines the order that clauses will be placed within generated SQL"
-  {:with 30
+  {:union 20
+   :union-all 25
+   :with 30
    :with-recursive 40
    :select 50
    :insert-into 60


seancorfield:
@jkk We just ran into this at World Singles, trying to create an ordered union query. This seems to be the simplest change that produces the correct ordering of clauses and doesn't seem to break anything else.

MichaelBlume:
Looks good to me.

(Note that until we get a release out, you *can* work around this by calling register-clause! at top-level from your own code at load-time.)

seancorfield:
Ah I wondered about that function. Thank you -- I'll try that as a workaround today since we only trip over that in one namespace at the moment.

seancorfield:
👍 Thank you!

MichaelBlume:
And released in 0.8.1, thanks =)

seancorfield:
Thank you!

PR #142, opened 2016-10-12
Don't parenthesize the subclauses of a UNION, UNION ALL, or INTERSECT clause.

### LINKED ISSUE #141: UNION clauses are parenthesized, which is not supported by SQLite
```
(deftest test-honeysql-union
  (is (= ["SELECT x FROM (SELECT x FROM abc UNION SELECT x FROM def) foo"]
         (sql/format {:select ['x]
                      :from (list [{:union (list
                                             {:select ['x]
                                              :from [:abc]}
                                             {:select ['x]
                                              :from [:def]})}
                                   :foo])}))))
```

fails: instead, honeysql produces

```
SELECT x FROM ((SELECT x FROM abc) UNION (SELECT x FROM def)) foo
```

This doesn't work with sqlite, and some opinions are that this is invalid SQL (though it works with MySQL) — I haven't read the standard myself.

See this bite others:

http://stackoverflow.com/questions/4653124/what-does-the-sql-standard-say-about-parentheses-in-sql-union-except-intersect-s#4653180
https://github.com/yiisoft/yii2/issues/9790
https://github.com/yiisoft/yii2/issues/8053
https://github.com/yiisoft/yii2/pull/8054

The only reason to parenthesize, AFAICT, is to avoid ambiguity when ordering and limiting one of the inner queries. (See #138/#140.)

But it utterly breaks SQLite, so I think that has priority!

rnewman:
Fixes #141.

As far as I've seen, the parens aren't allowed by the standard. Any use of the parens in engines that permit them can be solved by adding a simple subquery — see note in the test — so the best option is to simply remove them altogether.

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -560,13 +560,16 @@
   (str "WITH RECURSIVE " (comma-join (map cte->sql ctes))))
 
 (defmethod format-clause :union [[_ maps] _]
-  (string/join " UNION " (map to-sql maps)))
+  (binding [*subquery?* false]
+    (string/join " UNION " (map to-sql maps))))
 
 (defmethod format-clause :union-all [[_ maps] _]
-  (string/join " UNION ALL " (map to-sql maps)))
+  (binding [*subquery?* false]
+    (string/join " UNION ALL " (map to-sql maps))))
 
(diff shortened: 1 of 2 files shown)


MichaelBlume:
This looks good to me, anyone object?

PR #143, opened 2016-11-06
Update README example to match #142.

rnewman:


MichaelBlume:
Y'know, it actually really bothers me that we have all these examples in the README and no checks that they continue to be valid -- I don't suppose anyone knows of a way to automate running them?

(I guess we could have a readme_tests namespace and all the maintainers could remember that changes to that namespace/the README must be bidirectional)

PR #149, opened 2017-02-17
Fix operator/function calls in value positions

joodie:
Current version of the honeysql vector syntax does not correctly handle mathematical operators in value (comparison) positions. See the attached test and fix.


MichaelBlume:
Looks good to me, any objections?

joodie:
Is there anything I can do to get this change accepted?

MichaelBlume:
This, apparently, sorry.

joodie:
Thanks!

PR #150, opened 2017-02-18
Add helpers for :with and :with-recursive clauses

enaeher:
This allows writing:

```
(-> (with [:foo (-> (select :*)
                    (from :baz))])
    (select :quux)
    ...)
```

instead of

```
(-> {:with [[:foo (-> (select :*)
                      (from :baz))]]}
    (select :quux)
    ...)
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ## 0.8.3 In development
 
+* Add helpers for :with and :with-recursive clauses
+
 ## 0.8.2
 
 * Don't parenthesize the subclauses of a UNION, UNION ALL, or INTERSECT clause. (@rnewman)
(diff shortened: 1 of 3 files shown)


MichaelBlume:
Looks fine to me, any objections? Any reason we didn't do this in the first place?

emidln:
I think it might break static data. See my helpers from a talk I gave on honeysql. I included with-static, with-recursive, and with: https://github.com/emidln/honeysql-talk/blob/master/src/honeysql_talk/extra.clj#L10

enaeher:
Is there anything I can do to help this get merged? I don't understand the concern about static data. Is there a specific example I could add a test for?

MichaelBlume:
@emidln Could you clarify what you meant about static data?

MichaelBlume:
I'm merging, will release in a beta and see if anything comes up. Sorry for the delay.

MichaelBlume:
This is present in 0.9.0-beta2

PR #154, opened 2017-03-22
Fix using maps with different key orders in :values

### LINKED ISSUE #153: Array maps not supported?
When maps are clojure.lang.PersistentArrayMaps instead of clojure.lang.PersistentHashMaps, the parameter list can come out in the wrong order. Here's an example:

``` clojure
(require '[honeysql.core :as h] '[honeysql.helpers :as hh])

(-> (hh/insert-into :foo) 
    (hh/values [(array-map :a "a" :b "b" :c "c")  ;; note the different arg orders in the array-map calls
                (array-map :a "a" :c "c" :b "b")]) 
    (h/format))
;;=> ["INSERT INTO foo (a, b, c) VALUES (?, ?, ?), (?, ?, ?)" "a" "b" "c" "a" "c" "b"] ;; note the different orders

```

tomconnors:
Fixes #153. Expecting every map in the :values sequence to return the values in the same order when calling `vals` was the root cause.

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -541,10 +541,11 @@
   (if (sequential? (first values))
     (str "VALUES " (comma-join (for [x values]
                                  (str "(" (comma-join (map to-sql x)) ")"))))
-    (str
-      "(" (comma-join (map to-sql (keys (first values)))) ") VALUES "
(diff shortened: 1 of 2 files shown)


MichaelBlume:
Thanks!

PR #156, opened 2017-04-01
Require that defhelper arglist has two elements

### LINKED ISSUE #123: arglists metadata for helper functions is wrong
This is not super-important but the `:arglists` for functions like `merge-left-join` are wrong, which causes linters like [Eastwood](https://github.com/jonase/eastwood) to fail because it assumes you're calling these functions with the wrong number of arguments.

The arglists make it seem like the functions can only be called with 2 args:

``` clojure
(:arglists (meta #'honeysql.helpers/merge-left-join)) -> ([h clauses])
```

But in reality they can handle a variable number of args:

``` clojure
(h/merge-left-join :table [:= :a :b]) -> {:left-join (:table [:= :a :b])}
(h/merge-left-join {} :table [:= :a :b] [:= :c :d]) -> {:left-join (:table [:= :a :b] [:= :c :d])}
```

Not 100% sure what the right `:arglists` should be in this case but in the meantime I've had to manually alter the metadata to get my linter to stop complaining:

``` clojure
(alter-meta! #'honeysql.helpers/merge-left-join assoc :arglists '([m & clauses]))
```

MichaelBlume:
one for the map and one for varargs

Add & to metadata arglists, make them correct

Should fix #123

### DIFF
--- a/src/honeysql/helpers.cljc
+++ b/src/honeysql/helpers.cljc
@@ -15,9 +15,14 @@
 
 #?(:clj
     (defmacro defhelper [helper arglist & more]
-      (let [kw (keyword (name helper))]
+      (when-not (vector? arglist)
+        (throw (IllegalArgumentException. "arglist must be a vector")))
+      (when-not (= (count arglist) 2)
+        (throw (IllegalArgumentException. "arglist must have two entries, map and varargs")))
+      (let [kw (keyword (name helper))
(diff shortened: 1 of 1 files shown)


plexus:
I also ran into this using Eastwood. Any chance to get this merged? 

Thanks a lot!

Issue #177, opened 2017-08-10
Passing in a custom `defrecord` as a param results in `()`


yatesco:
I have a `defrecord MyCustomType` which returns a `java.sql.Timestamp` through extending `clojure.java.jdbc/ISQLValue`.

The following `(jdbc/query db ["select count(0) from some_table where some_column>?" (MyCustomType. ...)])` works but `(hsql/format {:where [:> :some_column (MyCustomType. ...)]})` returns `"WHERE some_column < ()"`

My assumption was that parameters were passed straight through but something seems to be (mis)interpreting instances of `MyCustomType`. 

Is there a magic protocol I need to implement to allow `honeysql` to understand `MyCustomType`?

Thanks!

MichaelBlume:
So we could have honeysql check whether an object `satisfies?` that protocol, though we'd have to make that check conditional on whether the jdbc lib appears to be present at all

yatesco:
I found implementing the [`honeysql.format/ToSql`](https://github.com/jkk/honeysql/blob/a707222d53ebe512ef47b6d771901cd9d75c4cc0/src/honeysql/format.cljc#L94) protocol works for now. 

Is that protocol a stable extension point?

MichaelBlume:
Yes, definitely

yatesco:
Issued a PR which updates the front page.

emidln:
I'm late here (as usual), but I was under the impression this was already solved more generically via `honeysql.format/Value` and its constructor `honeysql.format/value`. 

```(hsql/format {:where [:> :some_column (honeysql.format/value (MyCustomType. ...)]})``` would then return ```"WHERE some_column < ?"``` and defer to the underlying (JDBC) implementation for converting MyCustomType. This is much preferable to implement `ISQLValue` which is a well-known extension point rather than our own `ToSQL`, especially in applications that mix ways of building queries (inline SQL or HugSQL for relatively static queries; HoneySQL for more dynamic queries).

yatesco:
Hi @emidln. I think I will always argue for extending a protocol once rather than calling `honeysql.format/value` everywhere. Same for extending `ISQLValue` for date/times rather than the more painful `(java.sql.Timestamp. (.getTime....))`.

If it was Christmas :-) then  `honeysql` should remove the need to do anything special and pass unrecognised things to `clojure.java.jdbc`, or at the very least throw a meaningful exception.

emidln:
[...]

PR #179, opened 2017-08-22
Added doc to use custom datatypes re: #177

yatesco:
https://github.com/jkk/honeysql/issues/177


MichaelBlume:
It may still be worth putting a check for ISQL, especially once Clojure 1.9 drops and `satisfies?` is fast

PR #181, opened 2017-09-07
special handling for ISQLValue instances

### LINKED ISSUE #177: Passing in a custom `defrecord` as a param results in `()`

MichaelBlume:
fix #177

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -9,6 +9,26 @@
 
 ;;;;
 
+(def jdbc-present
+  #?(:clj
+     (try
+       (require 'clojure.java.jdbc)
+       true
+       (catch Exception e false))
+     :cljs false))
+
+(def jdbc-value?
+  (if jdbc-present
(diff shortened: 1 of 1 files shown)


seancorfield:
Would either @yatesco or @MichaelBlume be interested in providing an updated PR that passes the tests and will merge?

seancorfield:
Thanks, @yatesco I think I'll close this out and if someone wants to resurrect this at some future date, they can open a new issue or PR against the current codebase.

PR #186, opened 2017-11-07
Add :parameterizer :none option

arichiardi:
Sometimes you want to completely skip using parameters.

Postgres, in particular, complains with cryptic "syntax error at or near "$1"
whenever query parameters cannot be used for the query. An example would be
CREATE TABLE queries where the only parameters are the VARCHAR size.

This patch adds a :none option to :parameterizer which completely skips
parameter filling (and numbering), returning the SQL string only.

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -51,7 +51,8 @@
 
 (def ^:private parameterizers
   {:postgresql #(str "$" (swap! *all-param-counter* inc))
-   :jdbc (constantly "?")})
+   :jdbc (constantly "?")
+   :none #(str (last @*params*))})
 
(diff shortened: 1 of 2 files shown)


arichiardi:
Hello again 😄 I really really would like to add this if possible...it is blocking for me because the alternatives don't play well with `node-postgres`.

juvenn:
@arichiardi Awesome, just was looking for this as well, thank you!

Hi @MichaelBlume Any plan to cut a release to include this option?

MichaelBlume:
I've been meaning to cut a release for a while, haven't gotten around to updating the changelog in advance of a release. If somebody else wanted to do that, I could release sooner.

PR #189, opened 2017-11-16
Add :intersect to default-clause-priorities

kenfehling:


kenfehling:
I wasn't sure about the exact numbering, but after a bit of research it seemed that it should at least go before the UNION operators.

According to [the Oracle documentation](https://docs.oracle.com/database/121/SQLRF/queries004.htm#SQLRF52341) 

> To comply with emerging SQL standards, a future release of Oracle will give the INTERSECT operator greater precedence than the other set operators. Therefore, you should use parentheses to specify order of evaluation in queries that use the INTERSECT operator with other set operators.

Issue #203, opened 2018-01-29
Where with :and should remove `nil` clauses


arichiardi:
Simple example and maybe I am doing something wrong:

```clojure
cljs.user=> (-> (sqlh/where [:= :domain "ars"] [:= :entity "assar"]
                            (when false [:= :key "asr"])) 
                (sql/format  :parameterizer :postgresql))
["WHERE (domain = $1 AND entity = $2 AND NULL)" "ars" "assar"]
```

Probably the last `null` should not be generated.

MichaelBlume:
I think I agree that it'd be better if the NULL wasn't generated

arichiardi:
Solved by #205.

PR #204, opened 2018-02-15
Remove nil argument to where

visibletrap:
Try to addresses #203

### DIFF
--- a/src/honeysql/helpers.cljc
+++ b/src/honeysql/helpers.cljc
@@ -84,7 +84,7 @@
     [m pred logic-op]))
 
 (defn where [& args]
-  (let [[m pred] (prep-where args)]
+  (let [[m pred] (prep-where (remove nil? args))]
     (if (nil? pred)
       m
       (assoc m :where pred))))


MichaelBlume:
I've already merged this but if you wanted to submit another PR with a test case I'd appreciate it a lot and merge it right away.

visibletrap:
Thanks.

I need a your guide for a bit. What type of test do you prefer, a function focused test or high-level test? Should I add to `core_test.cljc` or create a `helpers_test.cljc`? Any test that you like which I could use as an example?

MichaelBlume:
I think something in format_test would be fine -- your example in #203 would make a good test case.

visibletrap:
First of all, I'm not the one requesting for this change. Just a random person pass by and found this easy enough for me to get started.

Second, I created a PR adding a test. But after seeing test examples in `format_test`, I kind of change my mind about the implementation that got merged. I'm not sure you would agree with me. And I don't want to trouble you too much so I present both options for you. Please choose to merge one of these and just close the other.

* #205 New implementation and its test (I prefer)
* #206 Only test for the current implementation

MichaelBlume:
Got it, thanks ^_^

PR #207, opened 2018-02-20
Add package.json for npm deployment

arichiardi:
This adds the necessary file that the maintainer needs in order to:

    npm publish

I decided to scope the package (same as a group in maven) so that things like
honeysql-postgres can use the same.

I am not including the `data_reader.clj` file at the moment because I haven't
actually had any problem not having it in `lumo` (admittedly I am not probably
noticing what the printing out looks like).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ pom.xml*
 .nrepl-port
 bin
 test/readme.clj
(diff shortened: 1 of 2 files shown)


arichiardi:
Branch name is weird, will open a new one.

PR #208, opened 2018-02-20
Add package.json for npm deployment

arichiardi:
This adds the necessary file that the maintainer needs in order to:
    
     npm publish

I decided to scope the package (same as a group in maven) so that things like
honeysql-postgres can use the same. You can see the tarball produced with:

    npm pack
    tar -xvlf honeysql-honeysql-v0.9.1.tgz

I am not including the `data_reader.clj` file at the moment because I haven't
actually had any problem not having it in `lumo` (admittedly I am not probably
noticing what the printing out looks like).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ pom.xml*
 .nrepl-port
 bin
 test/readme.clj
+/node_modules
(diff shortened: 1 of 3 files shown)


arichiardi:
@MichaelBlume After a bit of wrestling I got the Travis build to complete fine.
This is good to go and if you release a `0.9.2` on `npm` as well you will make me super happy :tada: :smile:

PR #209, opened 2018-03-09
Support register customized parameterizer

juvenn:
This extends on @arichiardi #186 idea, to support registering more customized parameterizer. For example in postgresql case, the value in sql string should be single-quoted.

Usage example:

```clojure
(require [honeysql.format :refer [format register-parameterizer])
(register-parameterizer :single-quote #(str \' % \'))
(format sql-map :parameterizer :single-quote)
```

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -56,6 +56,17 @@
    :jdbc (constantly "?")
    :none #(str (last @*params*))})
 
+(defn register-parameterizer
+  "Register f as a customized parameterizer.
+   E.g.:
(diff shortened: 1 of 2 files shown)


juvenn:
Sorry about failed build, I was encountering travis error "There was an error while trying to fetch the log." while trying to investigate how it failed.

I was able to pass all test cases locally via `lein test`. Will see if I can grab travis error log later.

arichiardi:
Oh this is a very helpful one!

MichaelBlume:
Nice work, thanks =)

juvenn:
Thanks Michael, would be glad to see a new release!

PR #214, opened 2018-05-15
Refactored parameterizer to use mutimethods

xlevus:
It was briefly mentioned in #honeysql on clojurians.slack.com that this
mechanism of extension uses a register method and an atom to support
extension when multimethods are used elsewhere.


arichiardi:
Nice!

PR #218, opened 2018-06-27
Use latest lumo for self-host testing

arichiardi:
This patch makes sure we use a fixed version (the latest) of lumo for self-host
testing and gets rid of the unnecessary nvm call now that we explicitly
install the latest libstdc++.


seancorfield:
👍

PR #222, opened 2018-08-13
Consider wrapping union sub-expressions in parenthesis

cddr:
Would you consider a change that wrapped sub-expressions in `union`, `union all`, and `intersection` statements. Otherwise it seems difficult to express e.g a `limit clause` in one of the sub-queries. I believe most databases will flatten it out at query optimization stage although I can't back that up with hard evidence.

### before
```
(sql/format {:union-all [{:select [:c1] :from [:t1]}
                         {:select [:c2] :from [:t2] :limit 1}]})

["SELECT c1 FROM t1 UNION ALL SELECT c2 FROM t2 LIMIT ?" 1]
```

### after
```
(sql/format {:union-all [{:select [:c1] :from [:t1]}
                         {:select [:c2] :from [:t2] :limit 1}]})
["(SELECT c1 FROM t1) UNION ALL (SELECT c2 FROM t2 LIMIT ?)" 1]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ Queries may be united within a :union or :union-all keyword:
 ```clojure
 (sql/format {:union [(-> (select :*) (from :foo))
                      (-> (select :*) (from :bar))]})
-=> ["SELECT * FROM foo UNION SELECT * FROM bar"]
+=> ["(SELECT * FROM foo) UNION (SELECT * FROM bar)"]
 ```
 
 Keywords that begin with `%` are interpreted as SQL function calls:
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -614,15 +614,18 @@
 
 (defmethod format-clause :union [[_ maps] _]
   (binding [*subquery?* false]
-    (string/join " UNION " (map to-sql maps))))
+    (string/join " UNION " (map (fn [x]
+                                  (str "(" (to-sql x) ")")) maps))))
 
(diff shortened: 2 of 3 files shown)


cddr:
Nice! I guess the examples in the README are somehow tested to ensure they match reality.

seancorfield:
See https://github.com/jkk/honeysql/issues/141 for why `union` is no longer parenthesized and https://github.com/jkk/honeysql/issues/138 for further discussion about how to handle `union`. TL;DR: we went down this path before and it broke things so we need to be very careful about changes like this. Happy to see extra work on this that analyzes the impact on a wide range of databases (so the modified queries need to actually work).

cddr:
I'm persuaded by #142 that this is not necessary. Don't see any benefit in introducing special cases to support non-standard behavior. Thanks.

```
  ;; UNION and INTERSECT subexpressions should not be parenthesized.
  ;; If you need to add more complex expressions, use a subquery like this:
  ;;   SELECT foo FROM bar1
  ;;   UNION
  ;;   SELECT foo FROM (SELECT foo FROM bar2 ORDER BY baz LIMIT 2)
  ;;   ORDER BY foo ASC
```

emidln:
Would it make sense to have a way of wrapping a clause in parens beyond a subselect? That way a user could opt in to a higher precedence.

seancorfield:
@emidln Feel free to open a new issue for that -- but I'm not convinced it would be worth the extra complexity it would add... so I think I'd want to see a PR as the basis of a discussion.

PR #225, opened 2018-08-30
Add Inlinable protocol for more sensible stringification of inline values

### LINKED ISSUE #224: An inline keyword should not retain its colon
Keywords are often used to represent constants in Clojure code, and they often correspond to ENUM values in the DB.

E.g. we use the ENUM value `unit` in Postgres and represent it as `:unit` in our Clojure code. However, this where-clause would not work because HoneySQL tries to interpret the keyword as a column:
```
[:= :some-column :unit]
```

Our current solution is to first convert the keyword to a string and then inline it, because it is not some unsafe thing that needs parametrization:
```
[:= :some-column (sql/inline (name :unit))]
```

However, I think the following syntax would be much easier to read and more sensible for any imaginable use case:
```
[:= :some-column (sql/inline :unit)]
```

Because of that, I propose that the way inline values are stringified should be improved, since there is no reason why Clojure's colon-representation of a keyword would make sense in SQL code.

vincent-dm:
The `SqlInline` record currently implements the `ToSql/to-sql` method by branching on the whether it is `nil` or not, and calling `str` on any non-nil value. Branching on type is not extensible and leads to suboptimal results (e.g. outputted SQL code represents inlined keywords with a starting colon).

My PR improves this by creating an new protocol `Inlinable` with an `inline-str` method which determines the stringification-behavior of inlined values by dispatching on their type.

It includes built-in implementations for `nil` and keywords, and a catch-all solution using `(str ...)`. It can also be extended by the user, so he can represent things like vectors in a way that suits his SQL flavor.

This PR fixes #224

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -1,6 +1,7 @@
 (ns honeysql.format
   (:refer-clojure :exclude [format])
-  (:require [honeysql.types :refer [call raw param param-name
+  (:require [honeysql.types :as types
+                            :refer [call raw param param-name inline-str
                                     #?@(:cljs [SqlCall SqlRaw SqlParam SqlArray SqlInline])]]
             [clojure.string :as string])
   #?(:clj (:import [honeysql.types SqlCall SqlRaw SqlParam SqlArray SqlInline])))
@@ -357,6 +358,18 @@
            (quote-identifier (second x))
            (to-sql (second x))))))
 
+(extend-protocol types/Inlinable
+  #?(:clj clojure.lang.Keyword
+     :cljs cljs.core/Keyword)
+  (inline-str [x]
+    (name x))
+  nil
+  (inline-str [_]
+    "NULL")
+  #?(:clj Object :cljs default)
+  (inline-str [x]
+    (str x)))
+
 (extend-protocol ToSql
   #?(:clj clojure.lang.Keyword
      :cljs cljs.core/Keyword)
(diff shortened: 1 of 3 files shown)


seancorfield:
Per my comment in #224 there's a bigger issue behind all of this involving both `inline` and `raw` but this moves the needle very slightly in a reasonable direction so I'm happy to include it for now (I actually think `inline :foo` should be an error and you should call `name` to convert a keyword to a string but your change doesn't break working code so...).

vincent-dm:
@seancorfield Thanks for the lightning-quick response!

When discussing my PR with my team, they brought up an interesting use case I didn't think of: namespaced keywords. In the "old" impl they would be inlined as `:foo/bar`, while with my PR they become `bar`. Maybe this breaks someone else's code?

Therefore, we could also leave out the protocol implementation for keywords: the default behavior then stays the same as before, but it is extensible through the protocol, so we can decide that we actually want inlined keywords represented using `(name ...)`.

I can ammend the PR if you want. Let me know.

seancorfield:
I'll leave it in place for now. I suspect both `inline` and `raw` will need protocol extension points for whatever solution I eventually come up with. If someone really is relying on on `:foo/bar` in their SQL code... well, I just can't imagine that in real world code to be honest...

...but feel free to update the PR if you want. I'm not in a rush to release 0.9.4 (I've updated the change log to include #224, #225, and a note about the changed behavior).

vincent-dm:
Okay, let's leave it as-is then. The namespaced use-case is very improbable indeed.

Issue #228, opened 2018-09-25
Merge-where produces an extra :and


arichiardi:
Hello folks!

I found that when I do `(sqlh/merge-where {} [])` I receive back a ` {:where [:and]}`.
It is probably a bug in the way `honeysql` handles the clauses.

Has anyone else experienced that?

Thank you :wave:

seancorfield:
I don't see that. Can you provide repro steps?
```
(! 619)-> clj -Sdeps '{:deps {honeysql {:mvn/version "RELEASE"}}}'
Clojure 1.9.0
user=> (require '[honeysql.core :as h] '[honeysql.helpers :refer :all])
WARNING: update already refers to: #'clojure.core/update in namespace: user, being replaced by: #'honeysql.helpers/update
nil
user=> (merge-where {} [])
{:where []}
user=> (-> (select :*) (from :table) (merge-where [:= :foo 42]))
{:select (:*), :from (:table), :where [:= :foo 42]}
user=> (h/format *1)
["SELECT * FROM table WHERE foo = ?" 42]
user=> 
```

arichiardi:
Yep sure, this is what I see:

```clojure
Clojure 1.10.0-alpha6
user=> (require '[honeysql.core :as h] '[honeysql.helpers :refer :all])
WARNING: update already refers to: #'clojure.core/update in namespace: user, being replaced by: #'honeysql.helpers/update
nil
user=>  (merge-where {:where [:and [:= 1 1]]} [])
{:where [:and [:and [:= 1 1]] []]}
```

seancorfield:
How did you create that inner hashmap? It looks wrong to me.

arichiardi:
Ah, that can be...let me see, thanks good point!

seancorfield:
```
user=> (where [:= 1 1])
{:where [:= 1 1]}
user=> (-> *1 (merge-where [:= :id 4]))
{:where [:and [:= 1 1] [:= :id 4]]}
```
This is why the helpers are provided -- a lot of people find the raw maps hard to get right. Best not to mix'n'match.

Can this be closed?

arichiardi:
I think so, thanks!

arichiardi:
Well ok, but this is data transformation and I though the semantics of `merge-where`, given that `merge` is in the name, would allow `nil` or empty in input...that's mainly where my confusion comes from. That's good to know anyways :wink:

seancorfield:
[...]

PR #229, opened 2018-09-25
Add minimal testing for #288

arichiardi:
I actually got some trouble with that commit so I am adding some testing here
and with time trying to understand the issue better.

### DIFF
--- a/test/honeysql/core_test.cljc
+++ b/test/honeysql/core_test.cljc
@@ -6,7 +6,7 @@
             [honeysql.helpers :refer [select modifiers from join left-join
                                       right-join full-join where group having
                                       order-by limit offset values columns
-                                      insert-into with]]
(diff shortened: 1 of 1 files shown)


seancorfield:
That test is bogus. You were trying to `(apply merge-where ...)` with no arguments. In the tests, you are calling `(merge-where … [])` which is definitely illegal and that's why the tests are failing with NPE.

seancorfield:
I assume you mean #228, not #288 by the way?

arichiardi:
Ah right, aren't the other two worth including though? I was coding at night so maybe I misinterpreted :smile:

Why closing the PR? It seems to me you are a bit too quick at closing things. At the end of the day I am trying to help no?

arichiardi:
I can open a new one with the correct number of the issue.

PR #231, opened 2018-10-01
The "AS" alias is no longer split before quoting

### LINKED ISSUE #216: AS should not split and quote
The `AS` alias should be accepted as-is, rather than being split and quoted. See PR #82 for more details.

gws:
Fixes #216.

:wave: Happy to change the little things if needed.


seancorfield:
:+1:

PR #233, opened 2018-12-30
Fix a typo in the Extensibility section

tlewin:
The current version doesn't work and returns a non-obvious error message:

```
interface honeysql.format.ToSql is not a protocol
```


seancorfield:
:+1:

PR #234, opened 2019-01-07
bind *parameterizer* in format-predicate

glittershark:
`format-predicate` throws an NPE if `*parameterizer*` is not bound - this
mirrors the binding form of `format` to allow supplying that as an
option.


seancorfield:
Will be in 0.9.5 (not sure when that will drop yet).

PR #237, opened 2019-02-09
Fix query generation when user.language is Turkish

### LINKED ISSUE #236: HoneySQL converts SQL keywords to uppercase using system Locale
`String.toUpperCase()` or `string/upper-case` for that matter converts the string to uppercase for the default system locale (system property `user.language`). Normally this still does what you'd expect but things like `inner join` get converted to `İNNER JOIN` (dot over the I) when user locale is Turkish. This predictably has bad consequences for people who like their SQL queries to work. The fix here is to use `String.toUpperCase(Locale/US)` instead which always converts things the way we'd expect.

I will submit a PR in a few minutes.

camsaul:
`String.toUpperCase()` (or `string/upper-case` for that matter) converts the string to uppercase for the default system locale (system property `user.language`). HoneySQL calls `string/upper-case` in a few places, for example when converting a key like `:join` to SQL keywords like `INNER JOIN`.

Normally this works the way you'd expect, but things like `inner join` get converted to `İNNER JOIN` (dot over the I) when user locale is Turkish. This predictably has bad consequences for people who like their SQL queries to work. The fix here is to use `String.toUpperCase(Locale/US)` instead which always converts things the way we'd expect.

Includes test. Fixes #236

### DIFF
--- a/test/honeysql/format_test.cljc
+++ b/test/honeysql/format_test.cljc
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [format])
   (:require [#?@(:clj [clojure.test :refer]
                  :cljs [cljs.test :refer-macros]) [deftest testing is are]]
+            honeysql.core
             [honeysql.types :as sql]
             [honeysql.format :refer
              [*allow-dashed-names?* quote-identifier format-clause format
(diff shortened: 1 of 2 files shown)


test/honeysql/format_test.cljc
```diff
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [format])
   (:require [#?@(:clj [clojure.test :refer]
                  :cljs [cljs.test :refer-macros]) [deftest testing is are]]
+            honeysql.core
```
camsaul: I launched CIDER in this namespace and it wouldn't load because a bunch of stuff is referring to `honeysql.core` (e.g. lines 229-231) and it wasn't required yet.

seancorfield:
That looks awesome! Great catch! i18n is such a complicated topic and a lot of Americans (and British, like me) folks forget about all sorts of edge cases around this. Thank you!

camsaul:
Hey @seancorfield, 

Do you have an expected timeline for the 0.9.5 release? We're heavy users of HoneySQL at [Metabase](https://github.com/metabase/metabase) and we have a couple of high-profile bugs our Turkish users are hitting. If you're not sure when a release would go out, we can patch downstream.

seancorfield:
It's been on my list for ages but I was hoping to get to a few other issues. I just pushed 0.9.5 to Clojars for you. I'll get to my other issues "later".

camsaul:
Thanks @seancorfield, I really appreciate it!

PR #239, opened 2019-03-25
Add support for MySQL REPLACE INTO

delenius:
This adds support for MySQL's "REPLACE INTO" syntax.
See: https://dev.mysql.com/doc/refman/5.5/en/replace.html

### DIFF
--- a/src/honeysql/format.cljc
+++ b/src/honeysql/format.cljc
@@ -216,6 +216,7 @@
    :union-all 45
    :select 50
    :insert-into 60
+   :replace-into 60
    :update 70
    :delete 75
    :delete-from 80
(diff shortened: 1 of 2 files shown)


seancorfield:
`REPLACE INTO` is non-standard (MySQL-specific) so it does not belong in HoneySQL. You can use HoneySQL's extension points to add it in your own code.

Note: there is a PostgreSQL-specific library of extensions to HoneySQL so this might be a good time to create a MySQL-specifci library and start out with adding `REPLACE INTO` that way.

delenius:
I tried that, but the problem I ran into was that I couldn't change `honeysql.format/default-clause-priorities`, so my "REPLACE INTO" ended up at the end of the SQL string. Is there a way around this?

seancorfield:
Use `honeysql.format/register-clause!` to add priorities for new clauses.

delenius:
Ah, yes, just found that :) That works, thanks.

seancorfield:
I've updated the README to include `register-clause!` (in the Extensibility section).

seancorfield:
(oops, typo'd the issue number in that commit message!)

PR #244, opened 2019-05-28
Bug: “Cannot determine type of empty array”

chriskilding:
Honeysql currently transforms an empty array into `ARRAY[]`. This fails because Postgres doesn’t know the type of the empty array, and Postgres emits the error “Cannot determine type of empty array”. We must change Honeysql to produce `’{}’` instead.

### DIFF
--- a/test/honeysql/format_test.cljc
+++ b/test/honeysql/format_test.cljc
@@ -85,6 +85,10 @@
 (deftest array-test
   (is (= (format {:insert-into :foo
                   :columns [:baz]
+                  :values [[(sql/array [])]]})
+         ["INSERT INTO foo (baz) VALUES ('{}')"]))
+  (is (= (format {:insert-into :foo
+                  :columns [:baz]
                   :values [[(sql/array [1 2 3 4])]]})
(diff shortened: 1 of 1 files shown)


chriskilding:
Looping in @callumcodes

seancorfield:
What other databases support array data types? (I know MySQL does not) Any change here needs to work with *all* databases that support array data types (otherwise this change should go into a DB-specific library, not the core HoneySQL one).

chriskilding:
Good point - off the top of my head, later versions of the ANSI SQL standard introduced array support, but it’s possible that not all DBs supported by honeysql in turn support this I suppose.

Does the core of Honeysql aim to support a particular ANSI SQL version?

seancorfield:
HoneySQL doesn't target a particular ANSI SQL version but it tries to ensure that it doesn't incorporate a behavior for one database that is at odds with what other databases need.

I wasn't able to find any examples of inserting an empty array value using ANSI syntax.

chriskilding:
In that case I’ll move this to an equivalent PR on the Postgres extension.

seancorfield:
Thanks!

PR #258, opened 2020-05-07
Update README.md

scott-silver:
typo


seancorfield:
Thank you for spotting that, and taking the time to submit a correction! Much appreciated.

PR #287, opened 2020-12-15
Mention in the README how to disable paremeterization

viebel:


seancorfield:
Perhaps not the best example to show since HoneySQL 1.0 does not correctly turn strings into valid SQL strings.

It does, however, work in 2.0:
```clojure
user=> (sql/format sqlmap {:inline true})
["SELECT a, b, c FROM foo WHERE f.a = 'baz'"]
```

I'll add the example to the v2 README and I'll merge this PR and add a caveat to the README about the 1.0 behavior.

PR #325, opened 2021-05-09
quoting for :%fun.col(s) syntax to match with [[:fun :col]]

zugnush:
Hi Sean,

I hope this is useful, to make quoting more consistent.

Regards
Neil

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -214,11 +214,13 @@
      (fn [fk _] (param-value (fk)))}))
 
 (defn- format-var [x & [opts]]
-  (let [c (name-_ x)]
+  (let [c (name x)]
     (cond (= \% (first c))
-          (let [[f & args] (str/split (subs c 1) #"\.")]
-            ;; TODO: this does not quote arguments -- does that matter?
-            [(str (upper-case f) "(" (str/join "," args) ")")])
+          (let [[f & args] (str/split (subs c 1) #"\.")
+                quoted-args (->> args
+                                 (map keyword)
+                                 (map format-entity))]
+            [(str (upper-case f) "(" (str/join "," quoted-args) ")")])
           (= \? (first c))
           (let [k (keyword (subs c 1))]
            (if *inline*
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -153,6 +153,7 @@
 
 (defn- namespace-_ [x] (some-> (namespace x) (str/replace "-" "_")))
 (defn- name-_      [x] (str/replace (name x) "-" "_"))
+(defn- -_          [x] (str/replace x "-" "_"))
 
 (defn- sqlize-value [x]
   (cond
@@ -217,15 +218,13 @@
   (let [c (name x)]
     (cond (= \% (first c))
           (let [[f & args] (str/split (subs c 1) #"\.")
-                quoted-args (->> args
-                                 (map keyword)
-                                 (map format-entity))]
-            [(str (upper-case f) "(" (str/join "," quoted-args) ")")])
+                quoted-args (map #(format-entity (keyword %) opts) args)]
+            [(str (upper-case (-_ f)) "(" (str/join ", " quoted-args) ")")])
           (= \? (first c))
-          (let [k (keyword (subs c 1))]
-           (if *inline*
-             [(sqlize-value (param-value k))]
-             ["?" (->param k)]))
+          (let [k (keyword (subs (-_ c) 1))]
+            (if *inline*
+              [(sqlize-value (param-value k))]
+              ["?" (->param k)]))
           :else
           [(format-entity x opts)])))
 
--- a/test/honey/sql/helpers_test.cljc
+++ b/test/honey/sql/helpers_test.cljc
@@ -866,27 +866,3 @@
            {:with [[:a]],
             :insert-into [[:quux [:x :y]]
                           {:select [:id], :from [:table]}]}))))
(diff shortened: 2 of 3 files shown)


src/honey/sql.cljc
```diff
@@ -214,11 +214,13 @@
      (fn [fk _] (param-value (fk)))}))
 
 (defn- format-var [x & [opts]]
-  (let [c (name-_ x)]
+  (let [c (name x)]
```
seancorfield, line 218: This change will break things unless you also change the `upper-case` call below to `sql-kw`. Consider:
```clojure
user=> (sql/format {:select [:%char-length.customer-name]})
["SELECT CHAR_LENGTH(customer_name)"]
```
At least one test should use kebab-case for both the function name and at least one column argument.

src/honey/sql.cljc
```diff
@@ -214,11 +214,13 @@
...
     (cond (= \% (first c))
-          (let [[f & args] (str/split (subs c 1) #"\.")]
-            ;; TODO: this does not quote arguments -- does that matter?
-            [(str (upper-case f) "(" (str/join "," args) ")")])
+          (let [[f & args] (str/split (subs c 1) #"\.")
+                quoted-args (->> args
+                                 (map keyword)
+                                 (map format-entity))]
```
seancorfield: `opts` needs to be passed through into `format-entity` here (as is done for the call near the end of this function and, ideally, tests need to also check that `:aliased` and `:drop-ns` options work correctly when passed through `format-var`. That also means that tests with qualified column names are needed, i.e., `f-o/b-r` as a column argument.

test/honey/sql/helpers_test.cljc
```diff
@@ -866,3 +866,27 @@
            {:with [[:a]],
             :insert-into [[:quux [:x :y]]
                           {:select [:id], :from [:table]}]}))))
+
+(deftest quoting-:%-syntax
```
seancorfield: Since this is really about testing the core `%` syntax, it should be in `sql_test.cljc` (and not use the helpers to construct the expression -- just `{:select [...]}` is sufficient.

src/honey/sql.cljc
seancorfield: Looking over the contexts that would set `:aliased` or `:drop-ns` to be `true`, I'm pretty certain that none of those would be legal for a function invocation so writing tests would be hard (short of actually testing the private function directly, which I try to avoid). I think just changing this to `(map #(format-entity % opts))` would be fine although I would prefer this expression to be `(map #(format-entity (keyword %) opts) args)` since it avoids the double-`map` and it's not really worth using the transduce arities here because there's likely only one or two arguments.

src/honey/sql.cljc
```diff
@@ -153,6 +153,7 @@
 
 (defn- namespace-_ [x] (some-> (namespace x) (str/replace "-" "_")))
 (defn- name-_      [x] (str/replace (name x) "-" "_"))
+(defn- -_          [x] (str/replace x "-" "_"))
```
seancorfield, line 156: Not needed -- see below.

src/honey/sql.cljc
```diff
@@ -214,16 +215,16 @@
...
 (defn- format-var [x & [opts]]
-  (let [c (name-_ x)]
+  (let [c (name x)]
     (cond (= \% (first c))
-          (let [[f & args] (str/split (subs c 1) #"\.")]
-            ;; TODO: this does not quote arguments -- does that matter?
-            [(str (upper-case f) "(" (str/join "," args) ")")])
+          (let [[f & args] (str/split (subs c 1) #"\.")
+                quoted-args (map #(format-entity (keyword %) opts) args)]
+            [(str (upper-case (-_ f)) "(" (str/join ", " quoted-args) ")")])
```
seancorfield, line 222: Instead of introducing a new function (see comment above), ~you can use `(sql-kw f)` as I suggested,~ (nope, I was wrong!)... instead of `(upper-case (-_ f))`, and that will be more in keeping with similar code elsewhere. I'm going to remove that function and just do the replace inline for now.

src/honey/sql.cljc
```diff
@@ -214,16 +215,16 @@
...
           (= \? (first c))
-          (let [k (keyword (subs c 1))]
-           (if *inline*
-             [(sqlize-value (param-value k))]
-             ["?" (->param k)]))
+          (let [k (keyword (subs (-_ c) 1))]
```
seancorfield, line 224: `-_` is not needed here. You've actually uncovered -- and almost fixed -- a bug that has been lurking for a while.

`:?foo-bar` should map to `{:params {:foo-bar 123}}` in the named parameters passed to `format` but the code currently calls `name-_` which would turn this into `:foo_bar` (which is actually wrong). Thank you for highlighting that bug!

It does, however, suggest to me that there's a deeper bug lurking here as well -- not covered in the new tests -- that I mentioned this in a previous review comment: the tests need to verify that qualified column names work properly (which they won't, so that's going to lead to more work needed here).

I think at this point I'm going to take the PR as-is and make the changes mentioned in the review, and beef up the tests and figure out how to fix the qualified column bug that I can now see exists in the code.

seancorfield (approved):
Thank you for getting the basic outline of the work in place via a PR. Much appreciated!

src/honey/sql.cljc
seancorfield: Just realized I was wrong about this -- `sql-kw` will replace `-` with space. Sorry.

PR #331, opened 2021-06-18
CI improvements

vemv:
These changes have in common that they intend to run tests and linting as nicely as possible.

Commit summary:

* Eastwood: also lint tests
* Make run-tests.sh a bit more robust
  * See e.g. https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
  * The `u` option is intentionally left out (as the script has some dynamism).
* Remove some redundant `->`s
* ~Correct various :arglists~
  * Often, these functions assume a "context" coming from a `->` call.
  * This context should be reflected in the :arglists - otherwise they stop being truthful and various tooling will display incorrect information.
  * This allows the newly enabled Eastwood linting to pass.

I checked in my fork that both GH Actions and Circle remain green.

Cheers - V

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,4 +25,4 @@ jobs:
           key: honeysql-{{ checksum "deps.edn" }}
       - run:
           name: Run all the tests
-          command: sh run-tests.sh all
+          command: ./run-tests.sh all
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: Clojure CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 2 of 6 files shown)


seancorfield:
Good catch on the redundant `->` and the misspelled `:arglist` (sic) instances.

That's a lot of changes all bundled together -- I'd prefer the CI changes separated out from the `:arglists` changes so they can be reviewed/discussed/applied separately. I have questions/comments on both sets of changes that will be easier to work through when they are two separate PRs (in particular, I think the optional first argument in the helpers should be named `query` and should never have an `_`).

Thanks.

vemv:
Sure, sounds good! I'll remove the :arglists changes from this PR and open a new one with those. Will rename the `_ctx` while I'm there.

seancorfield:
Can you explain the `sh run-tests.sh` -> `./run-tests.sh` change?

seancorfield:
I've incorporated as much of this as I'm willing to for now. Thank you. I'll try to remember to run similar changes into a couple of my other projects (I don't think any of them run GitHub Actions on PRs right now).

seancorfield:
I had to revert the `shellcheck` addition as it causes CI to fail -- because it rejects the `-E` option in `run-tests.sh`!

seancorfield:
And `-o` with `pipefail` is also illegal on GitHub Actions so I've removed that whole `set`. See the failed Actions builds here for details.

vemv:
All those failed because the mentioned `run-tests.sh -> ./run-tests.sh change` wasn't incorporated

Here is an explanation: https://stackoverflow.com/questions/54055549/linux-ubuntu-set-illegal-option-o-pipefail

seancorfield:
Interesting. I generally avoid `./` so that I don't have to worry about ensuring shell scripts are executable so that would be a hard habit to break I suspect.

PR #346, opened 2021-08-27
Test code blocks in docs with test-doc-blocks

### LINKED ISSUE #290: Switch from readme to test-doc-blocks
https://github.com/lread/honeysql/commit/dd0fa854c0e06ace40f5ecc269da3cc043eccb6f

Do this for both v1 and v2

lread:
Resolves #290

**Build**

New commands:
- `gen-doc-tests` - only regenerates tests if stale,
   use `clean` command to force regen
- `run-doc-tests` - calls gen-doc-tests then runs tests,
  accepts the same parameters as run-tests.
  Can specify `:platform`
    - `:cljs` - run tests under ClojureScript
    - otherwise Clojure where we can specify one of: `:1.9`
    `:1.10` `:master`

I'm not sure if my use of the `:platform` parameter jives with
your `:aliases` parameter used for `run-tests`.
Can adjust if you like.

Example usages:
```shell
clojure -T:build gen-doc-tests
clojure -T:build run-doc-tests
clojure -T:build run-doc-tests :aliases '[:cljs]'
clojure -T:build run-doc-tests :aliases '[:1.9]'
clojure -T:build run-doc-tests :aliases '[:1.10]'
clojure -T:build run-doc-tests :aliases '[:master]'
```

The `ci` command has been updated to generate and run doc tests for same
platforms as unit tests.

**Articles**

In addition to `README.md`, now testing doc blocks in all articles
under `doc` dir excepting `doc/operator-reference.md` which does not
have any runnable code blocks.

**Skipped**

Any code block that is intentionally not runnable has been marked to be
skipped via: `<!-- :test-doc-blocks/skip -->`.

**Consistency**

I noticed that some code blocks use REPL syntax:
```Clojure
user=> (+ 1 2 3)
6
```
and others use editor syntax:
```Clojure
(+ 1 2 3)
;;=> 6
```
some places also omit the comment for editor style:
```Clojure
(+ 1 2 3)
=> 6
```
All of this is just fine with test-doc-blocks.
I left the inconsistency as is, but can make a pass for consistency upon
request.

**HoneySQL state**

I noticed a code block that set the sql dialect was affecting other
tests. I simply restored the dialect to the default at the end of the
code block.

**Un-tweaked output**

Some code blocks had string output hand-tweaked for readability.
These have been adjusted to instead use `sql/format`'s `:pretty` option.
In some cases the output is not as readable as the hand-tweaked version.
I humbly suggest that perhaps `:pretty` output could perhaps be
improved (instead of having test-doc-blocks somehow adapt).

**Corrections**

There were very few code blocks that required fixing due to incorrect
output/code.  Please review the diffs carefully to make sure all is as
expected.

**ns requires adjustments**

Any specific case of `(ns my-ns (require [my-require :as a]))` is now
the REPL friendly `(require '[my-require :as a])`

Any missing required `requires` were added.

**ClojureScript :refer**

The HoneySQL docs seem to encourage the use of referred vars for
helpers. Although this has the con of overlaps with Clojure core vars,
it is also convenient for Clojure when using `:refer :all`.

ClojureScript does not support `:refer :all` and each var must be
specified in place of `:all`.

I have adjusted examples accordingly to work with both Clojure and
ClojureScript.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,12 +56,13 @@ From Clojure:
 From ClojureScript, we don't have `:refer :all`. If we want to use `:refer`, we have no choice but to be specific:
 <!-- {:test-doc-blocks/reader-cond :cljs} -->
 ```Clojure
+(refer-clojure :exclude '[filter for group-by into partition-by set update])
 (require '[honey.sql :as sql]
          '[honey.sql.helpers :refer [select select-distinct from
                                      join left-join right-join
                                      where for group-by having union
                                      order-by limit offset values columns
-                                     insert-into set composite
+                                     update insert-into set composite
                                      delete delete-from truncate] :as h]
          '[clojure.core :as c])
 ```
@@ -368,7 +369,7 @@ VALUES (?, (?, ?)), (?, (?, ?))
 Updates are possible too:
 
 ```clojure
-(-> (h/update :films)
+(-> (update :films)
     (set {:kind "dramatic"
            :watched [:+ :watched 1]})
     (where [:= :kind "drama"])
@@ -655,7 +656,7 @@ If `<table(s)>` and `<qualifier>` are both omitted, you may also omit the `[`..`
 (-> (select :foo.a)
     (from :foo)
     (where [:= :foo.a "baz"])
-    (h/for :update)
+    (for :update)
     (sql/format))
 => ["SELECT foo.a FROM foo WHERE foo.a = ? FOR UPDATE" "baz"]
 ```
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ for copying and pasting directly into your SQL tool of choice!
 
 ## Note on code samples
 
-All sample code in this README is automatically run as a unit test using
-[seancorfield/readme](https://github.com/seancorfield/readme).
+Sample code in this documentation is verified via
+[lread/test-doc-blocks](https://github.com/lread/test-doc-blocks).
 
 Some of these samples show pretty-printed SQL: HoneySQL 2.x supports `:pretty true` which inserts newlines between clauses in the generated SQL strings.
 
@@ -37,6 +37,8 @@ HoneySQL 1.x will continue to get critical security fixes but otherwise should b
 
 ## Usage
 
+From Clojure:
+<!-- {:test-doc-blocks/reader-cond :clj} -->
 ```clojure
 (refer-clojure :exclude '[filter for group-by into partition-by set update])
 (require '[honey.sql :as sql]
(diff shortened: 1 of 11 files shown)


lread:
Hiya @seancorfield, my first crack at it.
Happy to discuss/adjust. Lemme know!

seancorfield:
I would prefer that the gen test doc step be run as a task (in a separate process, as Eastwood and the tests are run -- and as readme is currently run) so that the `:build` alias doesn't need additional dependencies and `build.clj` doesn't have to require those additional libraries. That would mean moving that gen test doc code into a new directory and making it a `-main` so there could be a `:gen-test-doc` alias in `deps.edn` that could run it via `-M` -- then you can do `(bb/run-task opts [:gen-test-doc])`

seancorfield:
Not sure why `:platform` needs to be introduced given that the tests are already run under aliases for the cljs/clj-version variants?

lread:
> I would prefer that the gen test doc step be run as a task (in a separate process, as Eastwood and the tests are run -- and as readme is currently run) so that the `:build` alias doesn't need additional dependencies and `build.clj` doesn't have to require those additional libraries. That would mean moving that gen test doc code into a new directory and making it a `-main` so there could be a `:gen-test-doc` alias in `deps.edn` that could run it via `-M` -- then you can do `(bb/run-task opts [:gen-test-doc])`

Yep, can do.

lread:
> Not sure why `:platform` needs to be introduced given that the tests are already run under aliases for the cljs/clj-version variants?

Yeah... when your previous `build.clj` exposed `run-tests` on the command line (which it doesn't anymore?) I was a bit confused on how to use it.  I thought a `:platform` abstraction might be easier to understand and ran with that. 

Would you prefer a rename to `:alias`?  (I don't know if there is a need for an `:aliases` vector here.)

lread:
Oh while I'm thinking of it: not sure if it is important to you but `deps.edn` `:test` brings in the cognitect test runner dep.
This dep is not technically needed for `:cljs` testing but will be present because `:test` and `:cljs` are used together.

seancorfield:
> Would you prefer a rename to :alias? (I don't know if there is a need for an :aliases vector here.)

The existing code uses `:aliases` because that's what is used to build the basis for process execution and `run-task` operates in terms of a vector of alias names so let's keep it consistent and use `:aliases`

> not sure if it is important to you but deps.edn :test brings in the cognitect test runner dep.
This dep is not technically needed for :cljs testing but will be present because :test and :cljs are used together.

It's not important enough to want to change things around to avoid it (but it's a fair question to ask).

lread:
So @seancorfield, I think ae29147a2229193e13a4d0871c61be8e6ee6a41e is kind of what you were thinking, but details might need to be tweaked to your liking.

seancorfield:
That looks great @lread -- thank you so much for working on this!

lread:
> That looks great @lread -- thank you so much for working on this!

Thanks @seancorfield, it is always a pleasure with you!
Ping me if you find anything that needs love or could be improved in test-doc-blocks.

Issue #355, opened 2021-09-11
Alter table support for multiple alter options (Ex: multiple add column statements)


cyppan:
At least Mysql, Postgres and BigQuery support syntaxes like this:
```
ALTER TABLE table_name
ADD COLUMN ...,
ADD COLUMN ...,
...
```

AFAIU The DSL currently only supports one alter statement in the `:add-column` clause. It would be nice if the sql-helpers/add-column was additive instead.

seancorfield:
I'll give this some thought, thank you.

seancorfield:
This is supported as of 2.1.818 (I don't remember specifically adding support for this, but there are examples in the docs and those pass the tests):


```clojure
user=> (sql/format {:alter-table [:fruit
                                  {:add-column [:id :int [:not nil]]}
                                  {:drop-column :ident}]})
["ALTER TABLE fruit ADD COLUMN id INT NOT NULL, DROP COLUMN ident"]
user=> (sql/format {:alter-table [:fruit
                                  {:add-column [:id :int [:not nil]]}
                                  {:add-column [:name [:varchar 32]]}
                                  {:drop-column :ident}
                                  {:modify-column [:appearance :text]}]})
["ALTER TABLE fruit ADD COLUMN id INT NOT NULL, ADD COLUMN name VARCHAR(32), DROP COLUMN ident, MODIFY COLUMN appearance TEXT"]
user=> (sql/format {:alter-table [:fruit
                                  {:add-column [:id :int [:not nil] :if-not-exists]}
                                  {:drop-column [:if-exists :ident]}]})
["ALTER TABLE fruit ADD COLUMN IF NOT EXISTS id INT NOT NULL, DROP COLUMN IF EXISTS ident"]
```

PR #356, opened 2021-09-11
add support for multiple add/modify column statements

cyppan:
I'm not sure if it is something wanted and if I've done it the right way (`one-then-many-merge` looks odd maybe?)
My purpose was to be backward compatible for people who rely on the current add-column DSL structure.
Addresses #355


seancorfield:
I need to think about DSL changes in a more general way so I'll look at #355 but I'm not ready to take PRs for this sort of change.

PR #367, opened 2021-10-04
Allowing parameters as part of IS NULL/IS NOT NULL queries

dekelpilli:
My use case is with Postgres and json/jsonb operators (e.g. `[:= [:-> :jsonb-column-name "key-name"] nil]`). Currently, I need to build the keys into the query instead of using parameters.

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -1256,7 +1256,9 @@
                                    (if (= := op) " IS NULL" " IS NOT NULL"))
                               (cond-> nested
                                 (as-> s (str "(" s ")")))
-                              (vector))
+                              (vector)
+                              (into p1)
+                              (into p2))
                           (-> (str s1 " " (sql-kw op) " " s2)
                               (cond-> nested
                                 (as-> s (str "(" s ")")))


seancorfield:
Your example is unclear. Please provide a self-contained expression -- calling `format` -- that illustrates what is currently produced and what you believe should be produced. The proposed change makes no sense on just the basis of what you've written.

dekelpilli:
This is how it works with non-nil values, which is as expected:
```clj
(hsql/format
  {:select [[[:-> :obj "k"] :v]]
   :from   [:testing]
   :where  [:= [:-> :obj "check"] "desired-value"]})
=> ["SELECT obj -> ?  AS v FROM testing WHERE obj -> ? = ?" "k" "check" "desired-value"]
```
But with a nil comparison value, the `"check"` param is missing
```clj
(hsql/format
  {:select [[[:-> :obj "k"] :v]]
   :from   [:testing]
   :where  [:= [:-> :obj "check"] nil]})
=> ["SELECT obj -> ?  AS v FROM testing WHERE obj -> ? IS NULL" "k"]
```
This would give an error, as there are more param placeholders than params. The output I would expect instead is: 
```clj
["SELECT obj -> ?  AS v FROM testing WHERE obj -> ? IS NULL" "k" "check"]
```

Where `:->` is defined as follows
```clj
(hsql/register-op! :-> :variadic true)
```

seancorfield:
OK, so this is nothing to do with PostgreSQL or that `->` operator.

The following shows the problem:

```clojure
user=> (sql/format
  #_=>   {:select [[[:+ :obj 1] :v]]
  #_=>    :from   [:testing]
  #_=>    :where  [:= [:+ :obj 1] nil]})
["SELECT obj + ? AS v FROM testing WHERE (obj + ?) IS NULL" 1]
```

I'll work on this during the upcoming week. Thanks for catching this.

Issue #351, opened 2021-09-05
Cache generated SQL


seancorfield:
If `:cache` provided, cache SQL based on the set of keys in `:params`. Only works for named parameters.

seancorfield:
If `:cache` provided, assume a `clojure.core.wrapped` cache (atom), and perform a (dynamic) `lookup-or-miss` using the structure as a key, else format it (without `:cache`) and store the generated SQL.

There are good arguments for caching embedded parameters but also good arguments for disallowing them entirely. Perhaps default to disallow but provide an option to allow them?

seancorfield:
Fixed via #378

PR #378, opened 2022-01-08
first pass of caching formatter

seancorfield:
Addresses #351.

Needs more documentation.

Initial results suggest a speedup for simple queries of 2-3x.
Complex queries can see up to 20x speedup.


seancorfield:
The main issues to address in the documentation:
* Using embedded parameters will "work" but you'll get a cache entry for every different set of those embedded parameters.
* Using named parameters works better because you only get one cache entry (because parameters are substituted later).
* Users are responsible for cache tuning -- and should read about the different `core.cache` factories.

PR #393, opened 2022-03-19
Fix docstring of add-clause-before

frwdrik:
The result of adding a clause to a list in which the it already appears, is not to remove it from the list, but to move it to the end.


seancorfield:
Good catch, thank you!

PR #417, opened 2022-07-28
Change public-fns-have-clauses assertion to allow custom clauses

### LINKED ISSUE #416: Registering new clauses before loading honeysql helpers causes assertion error
If a library or an application registers sql clauses before honeysql helpers is loaded then this assertion will fail:

https://github.com/seancorfield/honeysql/blob/v2.2.891/src/honey/sql/helpers.cljc#L1031-L1036

To test this you can start up a repl using `clj -Sdeps '{:deps {com.github.seancorfield/honeysql {:mvn/version "2.2.891"}}}'` and then use the following code:

```clojure
(require '[honey.sql :as sql]
         '[clojure.string :as str])

(sql/register-clause!
 :on
 (fn [_ _] [])
 nil)

(require '[honey.sql.helpers :as sql.helpers])
```

And this will give you the following error:

```
Execution error (AssertionError) at honey.sql.helpers/eval1185 (helpers.cljc:1031).
Assert failed: (= (clojure.core/set (conj (clojure.core/deref (clojure.core/deref (var honey.sql/base-clause-order))) :composite :filter :lateral :over :within-group :upsert :generic-helper-variadic :generic-helper-unary)) (clojure.core/set (conj (map keyword (keys (ns-publics *ns*))) :nest :raw)))
```

corasaurus-hex:
Fixes #416

At the end of honey.sql.helpers there is an assert that confirms that all public functions have registered clauses. Formerly it was ensuring that the set of all public functions in the helpers namespace was identical to the set of currently valid clauses. This prevented registering new clauses before the helpers namespace was loaded.

This PR changes the assertion so that it compares against the original, unchanged set of clauses. This more accurately matches the intention of the assertion as well.

### DIFF
--- a/src/honey/sql/helpers.cljc
+++ b/src/honey/sql/helpers.cljc
@@ -1030,7 +1030,7 @@
        ;; ensure #295 stays true (all public functions have docstring):
        (assert (empty? (->> (ns-publics *ns*) (vals) (c/filter (comp not :doc meta))))))
      ;; ensure all public functions match clauses:
-     (assert (= (clojure.core/set (conj @@#'honey.sql/base-clause-order
+     (assert (= (clojure.core/set (conj @#'honey.sql/default-clause-order
                                         :composite :filter :lateral :over :within-group
                                         :upsert
                                         :generic-helper-variadic :generic-helper-unary))


seancorfield:
Thank you for spotting this and figuring out a solution!

PR #420, opened 2022-08-16
Support materialization control in WITH

### LINKED ISSUE #392: with materialized cte
### Discussed in https://github.com/seancorfield/honeysql/discussions/391

<div type='discussions-op-text'>

<sup>Originally posted by **k3nj1g** March  1, 2022</sup>
Can you help with register clause for materialized cte? For query like:

```
WITH foo AS MATERIALIZED (
  SELECT foo 
  FROM bar
)
```
</div>

See discussion for whether this should be a specialized `:with-materialized` or whether this should be the start of adding metadata support to provide hints on various clauses.

robhanlon22:
Adds an optional third value to `with` vectors, which can be the
following:

* `:materialized` -> SQL is `WITH cte AS MATERIALIZED (...)`
* `:not-materialized` -> SQL is `WITH cte AS NOT MATERIALIZED (...)`
* omitted or anything else -> SQL is `WITH cte AS (...)`

Note that materialization control is not available on WITH RECURSIVE
CTEs, so `format-with` was modified to take a third argument that
returns the `AS ...` separator, which is constantly `"AS"` for WITH
RECURSIVE, and obeys the aforementioned rules for non-recursive CTEs.

Resolves #392.

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -521,18 +521,20 @@
       (into [(str (format-entity (first x)) " " sql)] params))
     [(format-entity x)]))
 
-(defn- format-with [k xs]
+(defn- format-with [k xs as-fn]
   ;; TODO: a sequence of pairs -- X AS expr -- where X is either [entity expr]
   ;; or just entity, as far as I can tell...
   (let [[sqls params]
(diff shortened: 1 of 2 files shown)


robhanlon22:
thanks @seancorfield!

PR #428, opened 2022-09-08
Improve ClickHouse support

MawiraIke:
Thank you for this awesome library. I love using it to generate SQL statements.
I am using this library to generate ClickHouse database statements and although it is doable, it is in its current form some pain to generate complex queries (with a handful of clauses) due to missing clauses and minor differences between the syntax of currently supported dialects and ClickHouse. This leads to a lot of string concatenation which partly reduces the joy of using this library. To solve this, I have added more clauses, and statements and made some changes to the existing code which will make it smoother to generate ClickHouse statements with this library.
I am available to make any improvements/changes to this code. Thank you in advance.

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -43,26 +43,37 @@
 (def ^:private default-clause-order
   "The (default) order for known clauses. Can have items added and removed."
   [;; DDL comes first (these don't really have a precedence):
-   :alter-table :add-column :drop-column
+   :alter-table :add-column :drop-column :clear-column :comment-column
    :alter-column :modify-column :rename-column
-   :add-index :drop-index :rename-table
+   :alter-partition :alter-setting :alter-user :alter-quota :alter-role
+   :alter-settings-profile :alter-policy
+   :add-index :drop-index :rename-table :rename-type
    :create-table :create-table-as :with-columns
-   :create-view :create-materialized-view :create-extension
-   :drop-table :drop-view :drop-materialized-view :drop-extension
+   :create-view :create-materialized-view :create-live-view :create-window-view
+   :create-extension :create-database :create-function :create-user :create-role
+   :create-row-policy :create-quota :create-settings-profile :create-dictionary
+   :drop-table :drop-view :drop-materialized-view :drop-extension :drop-database
+   :drop-dictionary :drop-user :drop-role :drop-quota :drop-function :drop-row-policy
+   :drop-settings-profile
    :refresh-materialized-view
+   :watch :show :grant :revoke :explain :attach :detach :exists
    ;; then SQL clauses in priority order:
    :raw :nest :with :with-recursive :intersect :union :union-all :except :except-all
    :table
    :select :select-distinct :select-distinct-on :select-top :select-distinct-top
    :into :bulk-collect-into
-   :insert-into :update :delete :delete-from :truncate
-   :columns :set :from :using
+   :insert-into :update :delete :delete-from :delete-where :truncate :truncate-if-exists
+   :columns :set :from :using :on-cluster :to-name
+   :inner-engine :engine :watermark :allowed-lateness
+   :populate :clickhouse-comment :modify-comment :with-timeout :with-refresh :events
+   :sample :default-role
    :join-by
    :join :left-join :right-join :inner-join :outer-join :full-join
    :cross-join
-   :where :group-by :having
+   :prewhere
+   :where :group-by :limit-by :having
    :window :partition-by
-   :order-by :limit :offset :fetch :for :lock :values
+   :order-by :limit :offset :fetch :for :lock :into-outfile :clickhouse-format :values
    :on-conflict :on-constraint :do-nothing :do-update-set :on-duplicate-key-update
    :returning
    :with-data])
@@ -101,14 +112,15 @@
    (reduce-kv (fn [m k v]
                 (assoc m k (assoc v :dialect k)))
               {}
-              {:ansi      {:quote #(strop \" % \")}
-               :sqlserver {:quote #(strop \[ % \])}
-               :mysql     {:quote #(strop \` % \`)
-                           :clause-order-fn
-                           #(do
-                              (register-clause! :replace-into :insert-into :insert-into)
-                              (add-clause-before % :set :where))}
-               :oracle    {:quote #(strop \" % \") :as false}})))
+              {:ansi       {:quote #(strop \" % \")}
+               :sqlserver  {:quote #(strop \[ % \])}
+               :mysql      {:quote #(strop \` % \`)
+                            :clause-order-fn
+                            #(do
+                               (register-clause! :replace-into :insert-into :insert-into)
+                               (add-clause-before % :set :where))}
+               :oracle     {:quote #(strop \" % \") :as false}
+               :clickhouse {:quote #(strop "" % "")}})))
 
 ; should become defonce
 (def ^:private default-dialect (atom (:ansi @dialects)))
@@ -155,6 +167,11 @@
   "Helper to detect if SQL Server is the current dialect."
   []
   (= :sqlserver (:dialect *dialect*)))
+
+(defn- clickhouse?
+  "Helper to detect if Clickhouse is the current dialect."
+  []
+  (= :clickhouse (:dialect *dialect*)))
 
 ;; String.toUpperCase() or `str/upper-case` for that matter converts the
 ;; string to uppercase for the DEFAULT LOCALE. Normally this does what you'd
@@ -319,7 +336,9 @@
     (cond (= \% (first c))
           (let [[f & args] (str/split (subs c 1) #"\.")
                 quoted-args (map #(format-entity (keyword %) opts) args)]
-            [(str (upper-case (str/replace f "-" "_"))
+            [(str (if (clickhouse?)
+                    (str/replace f "-" "_")
+                    (upper-case (str/replace f "-" "_")))
                   "(" (str/join ", " quoted-args) ")")])
           (= \? (first c))
           (let [k (keyword (subs c 1))]
@@ -545,10 +564,17 @@
         (reduce-sql
          (map
           (fn [[x expr :as with]]
-            (let [[sql & params] (format-with-part x)
-                  [sql' & params'] (format-dsl expr)]
+            (let [[sql & params]   (format-with-part x)
+                  [sql' & params'] (if (and (clickhouse?)
+                                            (not (map? expr)))
+                                     (format-expr expr)
+                                     (format-dsl expr))]
               ;; according to docs, CTE should _always_ be wrapped:
-              (cond-> [(str sql " " (as-fn with) " " (str "(" sql' ")"))]
+              (cond-> [(if (clickhouse?)
+                         (if (map? expr)
+                           (str (str "(" sql' ")") " AS " sql)
+                           (str sql' " AS " sql))
+                         (str sql " " (as-fn with) " " (str "(" sql' ")")))]
                 params  (into params)
                 params' (into params'))))
           xs))]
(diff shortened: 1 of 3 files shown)


seancorfield:
As it stands, this is far too invasive as a set of changes.

First off, if the new clauses are dialect-specific, they should only be added dynamically when the ClickHouse dialect is selected. In addition, the corresponding helpers should be in an optional namespace so they don't further pollute `:refer :all` usage of `honey.sql.helpers`.

I'm also very concerned about the conditional logic introduced all over the place with the `clickhouse?` predicate -- that's going to make the core (ANSI) code harder to maintain and it will be too easy for me to break ClickHouse support in the future whenever I'm refactoring.

Looking at the tests, I can't tell what's ClickHouse-specific in them, vs what would normally be generated. It looks like the "stropping" for ClickHouse is an identity operation which seems... unusual? At a minimum, those tests need comments explaining what's different -- or better still have side-by-side assertions showing ANSI and ClickHouse results (i.e., so only the `:dialect` changes).

And, finally, all of this needs documentation: it adds a _huge_ number of extra clauses and they all need to be part of the documentation, even if they are confined to a specific dialect.

Overall, I don't think this has broad enough usage to warrant support out of the box, given that you can already register a new dialect that registers new clauses. It feels like it belongs in a companion `honeysql-clickhouse` library.

I'm happy to discuss the core changes with `clickhouse?` predicates in more detail to see if those can be handled without modifying the core logic, i.e., by providing different data structures in places where you think the default (ANSI) behavior isn't flexible enough.

MawiraIke:
Thank you @seancorfield for your detailed reply. 
About the dynamically added clauses. I can make a fix for this. Fair to note that I hadn't seen this in the code before but discovered it when merging your latest [commit](https://github.com/seancorfield/honeysql/commit/2f1d3ae8700df6e6f6c87183a4930f5ca65ff5f9#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R109) during PR creation.

About the conditionals. I tried to minimise that as much as possible but sometimes it was almost impossible to avoid it. Especially where the desired output was different. I could add more comments describing the difference but I doubt any more avoidance could be achieved, even after repeating the functions into new ones.

About the tests. I have included a lot more tests than I needed to. This was because I maintained all the tests I used when testing new clauses and even the existing ones that worked correctly as a sort of reference. I would be happy to add more details to the tests to show what's different.
Stropping is an identity because Clickhouse does not escape closing strop characters. There is no need to double the quote characters which leads to errors. Only strings as plain data are quoted. I have just realised that I could pass an empty map with no `quote` to [dialects](https://github.com/seancorfield/honeysql/blob/develop/src/honey/sql.cljc#L111) but for some reason, I think I got an error when I tried this earlier.

I would be keen to add documentation in docs that will expound on the new clauses more. During development, I tried to add sufficient documentation in the honey.sql.helpers namespace for every new helper added or changed.

The differences in core changes might be hard to circumvent without changing some of the core logic. Some examples:
1. Different ordering e.g in `WITH`.
`Ansi - WITH cte AS (SELECT * FROM example WHERE example_column = 0)`
`Clickhouse - WITH (SELECT * FROM example WHERE example-column = 0) AS cte`
This necessitates [this conditional](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R573)
In ClickHouse `WITH` expression passed to `WITH` can be a string. E.g 
`Ansi - (helpers/with [:abc "bcd"])` throws error but 
`ClickHouse - (h/with [:abc "bcd"])` produces `WITH 'bcd' AS abc` which is correct. This requires [this conditional](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R568)
2. Extra parenthesis needed e.g in `order-by`. E.g 
`Ansi - ORDER BY b DESC`
`ClickHouse - ORDER BY (b DESC)`
This requires [this conditional](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R773)
3. Extra flags like `if-not-exists` and `if-exists` in `Ansi` e.g `or-replace` in `clickhouse`. This necessitates [this conditional](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R981)
4. ClickHouse is case sensitive and does not take capitalization e.g in  `with columns`. I.e
`Ansi - (a STRING, b STRING, c STRING)`
`ClickHouse - (a String, b String, c String)`
This necessitates these conditionals [1](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R1052) and [2](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R1781).
5. `LIMIT` is different in ClickHouse. When the `offset` is provided, the `limit` and `offset` are not in parenthesis and also take some extra modifiers. E.g
`Ansi LIMIT (2, 10)`
`Clickhouse LIMIT 2, 10` and can also take extra modifiers i.e to become `LIMIT 2, 10 WITH TIES`
This necessitates [this conditional](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R1378)
6. Clickhouse also supports extra modifiers in `FETCH`. Currently, only `rows-only` is supported but in Clickhouse it can be `rows-with-ties`. This necessitates [this conditional](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R1396)
7. [This](https://github.com/seancorfield/honeysql/pull/428/commits/8eaa56f9b744a524747e7182e5dd2ade8490a0a4#diff-48b90b042f4ab4ec6e41e73fa591ef05dbdc1ad6a8d1e0d38f11c5bf06f3cba0R1039) can be written as a new function but would still need a conditional.
Going through the changes, I think those are the only changes to core logic that remain. Other changes are in form of new clauses which are ClickHouse specific.

seancorfield:
There's too much non-standard/non-ANSI stuff in this PR for me to feel comfortable accepting all those new clauses as core HoneySQL code -- I can't possibly maintain them and don't want code in the core where I might easily break ClickHouse support by making other changes.

Several of your conditionals can be avoided by providing alternative implementations of clauses -- which can be registered as part of your `set-dialect!` call: `:with`, `:limit`, `:offset`, `:fetch`. Note: HoneySQL does not produce `LIMIT (2, 10)` but `LIMIT 2 OFFSET 10`.

Your `ORDER BY` conditional seems unnecessary based on the ClickHouse docs: https://clickhouse.com/docs/en/sql-reference/statements/select/order-by -- no parentheses are mentioned there. I will also note that ClickHouse says they support the standard (ANSI) `WITH` syntax (`WITH identifier AS subexpression`) so there's no need for a special clause or conditional there either.

I'll give the `with-columns` issue some additional thought -- I've never been very happy with the auto-uppercasing of all but the first element, and there should be an escape hatch for that somehow -- but I'm going to close this PR out.

Feel free to ask questions in the #honeysql channel on the Clojurians Slack if you want help with implementing the ClickHouse dialect via `set-dialect!` and calls to `register-clause!` per the recent changes in 2.3.911.

seancorfield:
See #430 for the issue around as-is types etc in `with-columns`.

superstructor:
Thanks for your engagement and feedback on this @seancorfield 

Can you please confirm your position is that if we implement ClickHouse support as say, a `honeysql-clickhouse` library, that if we come across an inability for that library to support ClickHouse without changing code in `honeysql` itself that such a limitation will be treated by you as a bug or lacking a feature in extensibility (e.g. `set-dialect!`, `register-clause!` ) ?

MawiraIke:
> Your ORDER BY conditional seems unnecessary based on the ClickHouse docs: https://clickhouse.com/docs/en/sql-reference/statements/select/order-by -- no parentheses are mentioned there. I will also note that ClickHouse says they support the standard (ANSI) WITH syntax (WITH identifier AS subexpression) so there's no need for a special clause or conditional there either.

Few points to note on this in case you decide to add more flexibility. 
1. There seems to be multiple variants of `order-by` in clickhouse. E.g when doing `create-table` and passing order flag i.e `asc` or `desc`, clickhouse will fail if they are not enclosed in brackets. In my PR I had accommodated both `order-by` for `select` and `create-table` statements. 
E.g this works
``` sql
SELECT * FROM fake_table ORDER BY column_field ASC LIMIT 10
```

and this fails
``` sql
CREATE TABLE IF NOT EXISTS fake_table 
... 
ORDER BY column_field ASC
```

but works when you switch to
``` sql
CREATE TABLE IF NOT EXISTS fake_table 
... 
ORDER BY (column_field, ASC)
```

2. A similar situation is with `WITH`. If the expression is a subquery expression (for example a select statement), HoneySQL code will work as is fine but if the expression is not a subquery expression, i.e it is a string (which is not currently supported in HoneySQL), then the order in the current code will not work.
E.g this will run OK in Clickhouse

```sql
WITH '1970-01-01 00:00:00' as time ;; this is not supported in honeysql and throws an error (passing string as expr)
SELECT * FROM fake_table LIMIT 10
WHERE time_d = toDate(time)
```
But this will throw an error

```sql
WITH time as '1970-01-01 00:00:00'      ;; current honeysql order
SELECT * FROM fake_table LIMIT 10
WHERE time_d = toDate(time)
```

seancorfield:
@superstructor I would consider each separate issue on its own merits, on a case-by-case basis.

@MawiraIke Now that I've read more of the ClickHouse docs, I can see that the problem with your `CREATE TABLE` example is that ClickHouse's `ORDER BY` is part of the postlude for the DDL -- it is not an `ORDER BY` SQL clause -- which HoneySQL has no support for at all. The fact that you can add an `:order-by` clause to a `:create-table` clause is an accident of the implementation. There's no machinery in HoneySQL to specify `ENGINE` or things like `DEFAULT CHARSET` or any other post-create syntax. Tweaking the SQL clause is the wrong solution for DDL.

Thanks for clarifying ClickHouse's `WITH` syntax -- as noted earlier, I think your best bet is to use `register-clause!` for a ClickHouse-specific version of `:with` that supports what you need.

I will note one danger for dialects that register their own clauses (as the `:mysql` dialect currently does): any registered clauses are global and essentially become available across all dialects, so if you need to use HoneySQL for multiple dialects in a single application, generating SQL for multiple databases, you need to be careful that if you override a built-in clause, you do so in a way that is compatible with multiple dialects. The `:mysql` dialect example is "safe" because it is purely additive but what you will need to do with ClickHouse -- overriding some of the built-in clauses -- will need to be done with care.

seancorfield:
One final comment, since ClickHouse's `WITH expr AS ident` syntax does not cause ambiguities compared to the standard `WITH ident AS subquery` syntax, I'd be comfortable updating the built-in `:with` clause to support that because it couldn't break anyone's existing code -- and it would not need to be dialect-specific. Feel free to create a specific issue for that.

MawiraIke:
Thanks @seancorfield 
I have updated the built in `with` clause to add `with expr AS ident` syntax. I have created an issue #433

PR #432, opened 2022-09-22
Allow `with expr AS ident` syntax in WITH clause

MawiraIke:
This is to allow for ClickHouse WITH clause [variants](https://clickhouse.com/docs/en/sql-reference/statements/select/with).
See #428

### DIFF
--- a/doc/clause-reference.md
+++ b/doc/clause-reference.md
@@ -297,7 +297,8 @@ order they would appear in a valid SQL statement).
 
 These provide CTE support for SQL Server. The argument to
 `:with` (or `:with-recursive`) is a sequences of pairs, each of
-a result set name (or description) and a basic SQL statement.
+a result set name (or description) and either of; a basic SQL 
+statement, a string, a keyword or a symbol.
 The result set can either be a SQL entity (a simple name)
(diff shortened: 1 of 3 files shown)


seancorfield:
This still has references to ClickHouse. It should not. It also has no tests and no documentation updates. It also makes gratuitous indentation/whitespace changes.

It should be based purely on the types of the two parts of the `WITH` clause (whether they are `ident?`).

If you want to submit an updated PR with the above changes, feel free, but I'm happy to fix #433 myself when I get some time.

MawiraIke:
Okay, let me push an updated PR. I assumed you might have wanted to retain the current functionality for existing dialects. I.e throwing an error when the expression is not a subquery expression since the syntax is not supported anyway. I am updating the code to change the syntax if the expr is `ident?`

MawiraIke:
@seancorfield I have updated the code with the suggested changes. I am available to make any further changes. 
Also, I could not base it on `ident?` only and had to had `string?` because there is no simpler way to represent strings with spaces without further code changes.

seancorfield:
Thanks. I've kicked off a CI run and if it passes, I'll merge this in. Much appreciated!

PR #443, opened 2022-11-19
Improve type formatting logic in :cast function

duddlf23:
In the `:cast` function (special syntax), if the `type` is keyword or symbol, it can be quoted according to the format option.

```clojure
;; Example

(sql/format {:select [:foo [[:cast :bar :integer]]]} {:quoted true})
=> ["SELECT \"foo\", CAST(\"bar\" AS \"integer\")"]

(sql/format {:select [:foo [[:cast :bar :integer]]]} {:dialect :mysql})
=> ["SELECT `foo`, CAST(`bar` AS `integer`)"]
```

It is not valid that `type` to be quoted in the CAST( expr AS `type` ) syntax. Therefore, I modified the formatter of `:cast` to use `sql-kw` when the `type` is keyword or symbol. And I also added related tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -702,7 +702,7 @@ have a lot of function calls needed in code:
     (sql/format {:pretty true}))
 => ["
 INSERT INTO sample
-(location) VALUES (ST_SETSRID(ST_MAKEPOINT(?, ?), CAST(? AS integer)))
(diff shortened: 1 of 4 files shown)


seancorfield:
Thank you, looks good.

Issue #405, opened 2022-04-06
Support numbered parameters


seancorfield:
The PostgreSQL node client doesn't support `?` only `$N` parameters:

https://node-postgres.com/features/queries

stevenproctor:
Happy to help with this as needed

zerg000000:
@seancorfield Hi

I created a POC #444 for discussion. The changes invoked is a bit big and will break the extension api.
May I know what is the initial thoughts of how to implement this on your mind?

seancorfield:
I have a design in my head for a solution to this that breaks nothing but there are a couple of edge cases that require a bit more hammock time.

seancorfield:
The approach I plan to take with this is as follows:
* add an option to use numbered parameters (instead of positional)
* bind `*numbered* (atom [])` in `format` to store parameter values
* in the code that currently puts `?` into the SQL, put `$n` -- where `n` is `(count (swap! *numbered* conj <param>))` and have `(->numbered n)` for the actual parameter in the generated vector
* `->numbered` will work like `->param` but with the index (instead of a keyword) and based on `*numbered*` (instead of `*params*`)
* the existing code that derefs parameters in the vector (in `format`) will automatically work for this

This shouldn't break any extensions that are already using the low-level `format-*` API functions and it should also work with caching (with the same restrictions I think).

jaidetree:
That sounds like an ideal solution! Good thinking. I was able to make a function that works for now, though I don't know what-I-don't-know here but may be helpful until that gets implemented:

```clojure
(defn index-params
  ^:private
  [params]
  (let [unique-vals (vec (set params))
        param-index (->> unique-vals
                        (map-indexed #(vector %2 (inc %1)))
                        (into {}))]
    [unique-vals
     (->> params
          (map-indexed 
            (fn [idx value]
              [idx (str (get param-index value))]))
          (into {}))]))

(defn parse-honeysql-query
  ^:private
  "
  The hash-map shape this function takes is defined by the reducer
  in honeysql->pg
  "
  [{:keys [sql in-str escaping params-index index] :as state} char]
  (cond 
    (and (not in-str) (= char "?"))
    (assoc state
           :index (inc index)
           :sql (str sql "$" (get params-index index)))

    (and (not in-str) (or (= char "\"") (= char "'")))
    (assoc state 
           :in-str true
           :sql (str sql char))

    (and in-str (= char "\\"))
    (assoc state 
           :in-str false
           :escaping true
[...]

PR #444, opened 2022-11-27
[POC] Support numbered parameters

zerg000000:
POC #405

#### Approach

Change the internal presentation of sql from string to nested vector of sql fragments.
Convert back to SQL string only on public consumer API `sql/format`, `sql/format-expr`.

* Replace all `(str ...)` and `(str/join ...)` to `vector`/`interpose`.
* Replace hardcoded `"?"` to marker `:parameter-marker`
* Convert parameter marker to '?' or '$1' when convert sql fragments to string.

##### Good

* No impact on user that only use `helper` or `format`
* Minimum changes in current code base

##### Bad

* Breaking changes in extension API
* Internal Presentation become a forest of vectors

#### Checklist

- [x] workable implementation for discussion
- [x] Pass all tests
- [ ] Update documentation

#### Thoughts

* Align all internal/extend fn to output `[sqls params]`, all external fn to output `[sql param1 param2 ...]` may help to make code even more concise.
* Separate user api/extension api. e.g. `format-expr` separate into `format-expr` and `-format-expr`.
* Standardize the intermediate presentation format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -871,11 +871,11 @@ Or perhaps your database supports syntax like `a BETWIXT b AND c`, in which case
 ;; sequence of the arguments provided to it (so you can write any arity ops):
 (sql/register-fn! :betwixt
                   (fn [op [a b c]]
-                    (let [[sql-a & params-a] (sql/format-expr a)
-                          [sql-b & params-b] (sql/format-expr b)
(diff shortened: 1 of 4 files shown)


seancorfield:
Breaking the representation is not acceptable.

PR #452, opened 2023-01-07
Implement a basic SQLite dialect.

lnchan:
This dialect allows for supporting the same `REPLACE INTO` syntax supported by MySQL, without changing the quote format or other behaviors.  This allows the dialect to also be extended in the future with SQLite-specific syntax.

Alternatively, the `REPLACE INTO` syntax could be made to be used without any dialects.


seancorfield:
Thanks! I'll take a look at this when I get back from vacation in a week or so.

seancorfield:
Given that `REPLACE INTO` doesn't seem to conflict with any other dialect, I think your suggestion to just implement it in the "ANSI" core makes sense (since it also removes a weirdness caused by trying to mix MySQL and non-MySQL code, given that the MySQL dialect is side-effecting right now).

PR #457, opened 2023-02-08
Add section about `ORDER BY ... NULLS LAST` to 1.x differences dox

### LINKED ISSUE #449: Differences from Honey SQL 1 dox should mention things like `:desc-nulls-last` (2.x) vs `:desc :nulls-last` (1.x) in `:order-by`
In Honey SQL 1 if you wanted `ORDER BY ... DESC NULLS LAST` you would do something like

```clj
{:order-by [[:my-column :desc :nulls-last]]}
```

(two separate keywords)

In Honey SQL 2 this is 

```clj
{:order-by [[:my-column :desc-nulls-last]]}
```

This tripped me up when converting some old code that used Honey SQL 1 today -- it would be good to mention this in the "Differences from Honey SQL 1" documentation

camsaul:
Resolves #449

### DIFF
--- a/doc/differences-from-1-x.md
+++ b/doc/differences-from-1-x.md
@@ -201,7 +201,6 @@ you needed to pass `:nulls-last` as a separate keyword, after `:asc` or `:desc`:
 
 ```clj
 {:order-by [[:my-column :desc :nulls-last]]}
-;; =>
 ```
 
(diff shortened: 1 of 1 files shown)


seancorfield:
@camsaul Good catch. Can you remove the stray `;; =>` so that the doc tests don't fail?

camsaul:
Oops. Fixed

PR #468, opened 2023-02-26
Allow multiple expressions in special :order-by

### LINKED ISSUE #465: Can't specify multiple columns in the `ORDER BY` clause in aggregate functions
The issue is about PostgreSQL but might be relevant to other DBs.

An example SQL syntax would be:
```sql
SELECT ARRAY_AGG(id ORDER BY degree ASC, amount DESC)
FROM my_table;
```

The [aggregate functions](https://www.postgresql.org/docs/15/functions-aggregate.html) documentation says:
> The aggregate functions `array_agg`, `json_agg`, `jsonb_agg`, `json_object_agg`, `jsonb_object_agg`, `string_agg`, and `xmlagg`, as well as similar user-defined aggregate functions, produce meaningfully different result values depending on the order of the input values. This ordering is unspecified by default, but can be controlled by writing an `ORDER BY` clause within the aggregate call, as shown in [Section 4.2.7](https://www.postgresql.org/docs/15/sql-expressions.html#SYNTAX-AGGREGATES).

That linked section 4.2.7 says:
> The `order_by_clause` [which is used within aggregate functions] has the same syntax as for a query-level `ORDER BY` clause, as described in [Section 7.5](https://www.postgresql.org/docs/15/queries-order.html), except that its expressions are always just expressions and cannot be output-column names or numbers.

As that last section 7.5, as expected, describes that an `ORDER BY` clause can have multiple columns specified, each with its own order, as in `ORDER BY col1, col2 ASC, col3 DESC`.

p-himik:
Fixes #465 

Starting as a PoC, not sure whether further changes are needed (docs, tests,...).

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -1513,9 +1513,9 @@
       (let [[sql & params] (format-expr x {:nested true})]
         (into [(str "NOT " sql)] params)))
     :order-by
-    (fn [k [e q]]
+    (fn [k [e & qs]]
       (let [[sql-e & params-e] (format-expr e)
-            [sql-q & params-q] (format-dsl {k [q]})]
+            [sql-q & params-q] (format-dsl {k qs})]
         (-> [(str sql-e " " sql-q)]
             (into params-e)
             (into params-q))))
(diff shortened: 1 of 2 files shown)


seancorfield:
This looks good. Thank you!

PR #469, opened 2023-02-26
Add optional type argument to the :array special

### LINKED ISSUE #464: No way to cast to an array of a specific type without using `:raw`
An example SQL syntax from https://www.postgresql.org/docs/current/arrays.html#ARRAYS-IO:
```sql
SELECT f1[1][-2][3] AS e1, f1[1][-1][5] AS e2
 FROM (SELECT '[1:1][-2:-1][3:5]={{{1,2,3},{4,5,6}}}'::int[] AS f1) AS ss;
```
An alternative to `{...}::int[]` is `CAST(... AS int[])`.

I usually need it with empty arrays because otherwise JDBC or something else can't figure out what type to actually use which results in an exception.
Another use case is when you have an enum and want to pass an array with values from that enum to the DB.

My current workaround is to use `[:cast [:array the-array] [:raw "integer[]"]`.

p-himik:
Fixes #464 

`:array` seems like a natural place for the type given that the issue is specifically about casting arrays.
If the fix is reasonable, some doc changes are also needed.

### DIFF
--- a/doc/special-syntax.md
+++ b/doc/special-syntax.md
@@ -8,15 +8,19 @@ The first group are used for SQL expressions. The second (last group) are used p
 
 ## array
 
-Accepts a single argument, which is expected to evaluate to
-a sequence, and produces `ARRAY[?, ?, ..]` for the elements
-of that sequence (as SQL parameters):
+Accepts a single argument, which is expected to evaluate to a sequence,
+with an optional second argument specifying the type of the array,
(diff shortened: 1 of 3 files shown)


seancorfield:
Looks like a reasonable change. Could you update the PR with a first pass at updating the docs for the `:array` special syntax?

p-himik:
Done!

PR #470, opened 2023-02-26
Allow single kw in :order-by and other clauses

### LINKED ISSUE #467: Inconsistent requirements for `:order-by` and `:group-by`
Clauses like `:select`, `:from`, `:returning`, `:where`, `:having`, etc. accept a single keyword (nonsensical SQL just for demonstration):
```clj
=> (honey.sql/format {:select    :a
                      :from      :b
                      :returning :d
                      :where     :e
                      :having    :f
                      :window    [:w {:partition-by :name
                                      :order-by     [:id]}]})
["SELECT a FROM b WHERE e HAVING f WINDOW w AS (PARTITION BY name ORDER BY id ASC) RETURNING d"]
```

And the README [mentions it](https://github.com/seancorfield/honeysql/blob/de0adf56ef359615d9459dcbc4efe5037eb35288/README.md?plain=1#L595), although quite late, alter having already used `:select` et al. with a vector of a single keyword.

Whereas `:order-by` and `:group-by`, when passed a single keyword, make `format` throw:
```clj
=> (honey.sql/format {:order-by :a})
Execution error (IllegalArgumentException) at honey.sql/reduce-sql (sql.cljc:454).
Don't know how to create ISeq from: clojure.lang.Keyword

=> (honey.sql/format {:group-by :a})
Execution error (ExceptionInfo) at honey.sql/format-expr-list (sql.cljc:480).
format-expr-list expects a sequence of expressions, found: class clojure.lang.Keyword

=> (honey.sql/format {:window [:w {:order-by :id}]})
Execution error (IllegalArgumentException) at honey.sql/reduce-sql (sql.cljc:454).
Don't know how to create ISeq from: clojure.lang.Keyword
```

Wouldn't it make sense to make `:order-by` and `:group-by` clauses consistent with the others mentioned?

p-himik:
Fixes #467 

Also fixes not being able to use helpers to merge into stuff like `{:select :id}` - e.g. `(hh/select {:select :id} :x)` would throw.

I would also update README so it mentions the "plain" syntax earlier and uses it wherever possible.

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -209,6 +209,9 @@
                       {:symbol x
                        :failure (str t)})))))
 
+(defn- ensure-sequential [xs]
+  (if (sequential? xs) xs [xs]))
+
 (defn format-entity
   "Given a simple SQL entity (a keyword or symbol -- or string),
   return the equivalent SQL fragment (as a string -- no parameters).
@@ -549,7 +552,7 @@
     (-> [sql'] (into params) (into params'))))
 
 (defn- format-select-into [k xs]
-  (let [[v e] (if (sequential? xs) xs [xs])
+  (let [[v e] (ensure-sequential xs)
         [sql & params] (when e (format-expr e))]
     (into [(str (sql-kw k) " " (format-entity v)
                 (when sql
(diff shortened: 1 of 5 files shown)


seancorfield:
I'll think about the documentation aspect. This was a new "feature" in v2.x that fell out of an implementation detail for the most part. I'll create an issue for doc updates.

PR #473, opened 2023-03-01
pg-ops: add docs

holyjak:
I always struggle to remember which operator does what so it would be super helpful to have that docs directly in the ns.

Also, I find examples of usage very useful but hard to come by so I added some.


seancorfield:
Good stuff! Thank you!

PR #490, opened 2023-05-20
improve on-conflict docstring

holyjak:
It was unclear to me what the "hash map" where-condition meant so I tried to clarify.


seancorfield:
Thank you!

seancorfield:
FWIW, the hash map can be "any" SQL clause but I believe only `:where` or `:on-constraint` are valid SQL.

PR #491, opened 2023-06-02
honey.sql/upper-case for clojure-clr

Dangercoder:
Added honey.sql/upper-case for clojure-clr. Tested via the repl. 

Best,
Dangercoder


seancorfield:
Thank you! Weird casing `en-Us` 👀

PR #502, opened 2023-08-18
Example in README for custom function usage in select

markbastian:
Adding an example to the README.md file to clarify how custom functions are used in select statements along with their corresponding column alias.


seancorfield:
Thanks for that @markbastian !

PR #508, opened 2023-10-16
Unhashable values

jcf:


seancorfield:
Fixed on **develop** using a slightly different approach.

PR #514, opened 2023-11-15
Proposed opt `quoted-when`

kranstrom:
Hello @seancorfield 

Proposing a new option, `quoted-when`, to allow overriding the default (opinionated) quoting when quoting has been disabled.

This can be useful, for example, if a user wants to selectively quote fields like only quoting reserved words.

Welcome to any feedback and appreciate all the great work on honeysql.

Best,

### DIFF
--- a/doc/options.md
+++ b/doc/options.md
@@ -22,6 +22,7 @@ All options may be omitted. The default behavior of each option is described in 
 * `:numbered` -- a Boolean indicating whether to generate numbered placeholders in the generated SQL (`$1`, `$2`, etc) or positional placeholders (`?`); the default is `false` (positional placeholders); this option was added in 2.4.962,
 * `:params` -- a hash map providing values for named parameters, identified by names (keywords or symbols) that start with `?` in the DSL; the default is that any such named parameters will have `nil` values,
 * `:quoted` -- a Boolean indicating whether or not to quote (strop) SQL entity names (table and column names); the default is `nil` -- alphanumeric SQL entity names are not quoted but (as of 2.3.928) "unusual" SQL entity names are quoted; a `false` value turns off all quoting,
+* `:quoted-when` -- a function overriding the default behavior of when to quote entities when `:quoted false` (default). Function should take single argument - the entity/part being quoted - and return a boolean indicator whether to quote it or not.
 * `:quoted-snake` -- a Boolean indicating whether or not quoted and string SQL entity names should have `-` replaced by `_`; the default is `false` -- quoted and string SQL entity names are left exactly as-is,
 * `:values-default-columns` -- a sequence of column names that should have `DEFAULT` values instead of `NULL` values if used in a `VALUES` clause with no associated matching value in the hash maps passed in; the default behavior is for such missing columns to be given `NULL` values.
 
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -120,6 +120,7 @@
 ; should become defonce
 (def ^:private default-dialect (atom (:ansi @dialects)))
 (def ^:private default-quoted (atom nil))
+(def ^:private default-quoted-when (atom nil))
 (def ^:private default-quoted-snake (atom nil))
 (def ^:private default-inline (atom nil))
 (def ^:private default-checking (atom :none))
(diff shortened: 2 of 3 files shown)


p-himik:
What problem does it solve? Not a subjective "I want fewer quotes" but something objective, like "this particular DB doesn't like some things being quoted".

Also, have you considered something like this that already exists, where quoting (and some replacements) is opt-out?
```clojure
(sql/format {:select [:a :a-b :'a-b]}
            {:quoted true})
=> ["SELECT \"a\", \"a-b\", a-b"]
```

nharsch:
In our _specific_ case, we want to extend the set of reserved words and special characters that would trip a "safety quote". For instance, we'd like to quote all idents that start with a non alpha character, and any DB specific reserved words. 

I think this is generically a good feature to have.

p-himik:
Yeah, but why not just quote everything? What are the downsides that would make this feature worth it?

seancorfield:
As @p-himik asks: what is the problem with the current "quote everything" option that needs solving for your use cases?

I would rather not have an arbitrary predicate for this, given `:quoted` and `:quoted-snake` are essentially both Boolean (or `nil`).

@nharsch Would your use case be satisfied by being able to specify a different regex for the current quote-if-not-alphanumeric logic? (or perhaps an _additional_ regex that was a positive test, in addition to the current "negative" test?)

As a point of procedure, I would _prefer_ discussions of issues happen first _before_ anyone moves to a PR. Normally, I'd just close out a PR that doesn't already have an issue that it addresses, but I'll leave this open _as a discussion_ to hear answers from @kranstrom and @nharsch (but I will close this in a few days, regardless: if there is an actual problem to be solved, an issue can be opened to track that).

nharsch:
> Would your use case be satisfied by being able to specify a different regex for the current quote-if-not-alphanumeric logic? (or perhaps an additional regex that was a positive test, in addition to the current "negative" test?)

I believe so, in fact @kranstrom is looking into that. 

> As a point of procedure, I would prefer discussions of issues happen first before anyone moves to a PR.

@seancorfield For sure. Should we open those in github discussions, and/or on clojurians slack?

kranstrom:
Thanks for all the great feedback here.

@seancorfield 

> I would rather not have an arbitrary predicate for this

I get this. The thought behind this PR was essentially building on top the existing default (don't quote anything), but honeysql will still quote based on this regex `^[A-Za-z0-9_]_+`.

To your point, the implementation in this PR could (if needed at all) be different, but the thought was essentially extending upon this selectively quoting behavior.

To answer the "why?" question, the thought is to quote only when needed (reserved words, field formatting, etc.). In doing so, the honeysql syntax can be written with less consideration for case. It would also assist us in this translation from sources that we are pulling data from (and can't control), with targets where we can. Hopefully this helps some.

All this to say is that this was a proposal. Definitely appreciate the `'` suggestion by @p-himik  as I had missed this.

If this is something that could be beneficial, but using a different implementation - I can adjust based on recommendations. If we'd rather use existing functionality and not do this at all, no hard feelings here.

Thanks everyone!

kranstrom:
@seancorfield  will be sure to open issues in the future before proposing a PR. Thanks.

seancorfield:
> I get this. The thought behind this PR was essentially building on top the existing default (don't quote anything), but honeysql will still quote based on this regex `^[A-Za-z0-9_]_+`.

It only does that if `:quoted nil`. If you explicitly specify `:quoted false` it will turn quoting off. The behavior is:
* `:quoted true` -- always quote using the dialect's stropping
* `:quoted nil` -- default: only quote non-alphanumeric entities
* `:quoted false` -- do not quote anything

I gather from your response that the **problem** you want to solve here is that:
* "always quoting" makes things case-sensitive and you don't want that (because you're working from existing SQL that would fail if entities were quoted -- because it already doesn't match the case of tables/columns in the database)
* and the existing "quote non-alphanumerics" isn't enough because you have columns in the existing SQL that are reserved words in your new SQL database (that presumably weren't in the old SQL database)
* and you can't use the "never quote" approach either due to the reserved words issue

Is that an accurate expression of the **problem**?

kranstrom:
@seancorfield that is accurate - thanks for writing it out so clearly.

Another example that is applicable is that database technologies like SQL Server do not support the first character in a field name being a number _unless_ the field is quoted. It's an example where the "quote non-alphanumerics" may have some room for improvements or addressing the problem description could be helpful.

Really appreciate your feedback on this one - thanks again @seancorfield

seancorfield:
@kranstrom it turns out that some databases are perfectly happy with entity names that start with a digit (and, in fact, are all digits and underscores) so I don't want to quote `0abc` by default because it isn't needed for everything.

So that leads me to think that we need dialect-based quoting rules, possibly with overrides. I'll have to give that some thought, and at this point I'm going to close this PR (there are conflicts now), and I'll create a new issue describing the problem better.

PR #517, opened 2023-12-08
Implement CREATE INDEX

### LINKED ISSUE #348: Add some sort of support for CREATE INDEX
There's no standard for this aspect of DDL it seems, but here's:
* PostgreSQL https://www.postgresql.org/docs/current/sql-createindex.html
* SQL Server https://docs.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver15

They're different (of course) but have some common elements. It's not clear what HoneySQL syntax should be for this because there are some words that come between `CREATE` and `INDEX`, some that come between `INDEX` and the name (which is optional in PG but not in MS! Argh!) and `WITH`/`WHERE` are in different orders between PG and MS too. Nasty.

dancek:
Fixes #348. Implemented based on the discussion in that issue.

### DIFF
--- a/doc/clause-reference.md
+++ b/doc/clause-reference.md
@@ -98,7 +98,7 @@ names: the "from" and the "to" names.
 
 > Note: `:modify-column` is MySQL-specific and should be considered legacy and deprecated. `:alter-column` will produce `MODIFY COLUMN` when the MySQL dialect is selected.
 
-### add-index, drop-index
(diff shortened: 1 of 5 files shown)


dancek:
I'll be happy to fix any issues in this PR, whether logical or stylistic!

seancorfield:
This looks great! Thank you!

PR #546, opened 2024-09-27
Hodgepodge of optimizations

alexander-yakushev:
Following #545, here's a bunch of further improvements to HoneySQL efficiency. Commits are meaningful, I've included the explanations in the commit descriptions.

Benchmark I've used:

```clj
(format {:select [[[:min :recent_views.user_id] :user_id]
                  :model
                  :model_id
                  [[:max [:coalesce :d.view_count :t.view_count]] :cnt]
                  [:%max.timestamp :max_ts]]
         :group-by  [:model :model_id]
         :where     [:and
                     [:= :context "view"]
                     [:in :model #{"dashboard" "table"}]]
         :order-by  [[:max_ts :desc] [:model :desc]]
         :limit     10
         :left-join [[:report_dashboard :d]
                     [:and
                      [:= :model "dashboard"]
                      [:= :d.id :model_id]]
                     [:metabase_table :t]
                     [:and
                      [:= :model "table"]
                      [:= :t.id :model_id]]]}
        {:dialect :ansi})
```

```
Before:
Time per call: 50.31 us   Alloc per call: 121,104b
Time per call: 45.95 us   Alloc per call: 121,171b
Time per call: 43.92 us   Alloc per call: 121,104b

After:
Time per call: 35.65 us   Alloc per call: 82,392b
Time per call: 35.39 us   Alloc per call: 82,392b
Time per call: 35.25 us   Alloc per call: 82,392b
```

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -280,7 +280,7 @@
 
   Handles quoting, splitting at / or ., replacing - with _ etc."
   [e & [{:keys [aliased drop-ns]}]]
-  (let [e           (if (and aliased (keyword? e) (= \' (first (name e))))
+  (let [e           (if (and aliased (keyword? e) (str/starts-with? (name e) "'"))
                       ;; #497 quoted alias support (should behave like string)
                       (subs (name e) 1)
                       e)
@@ -350,7 +350,7 @@
     (let [n (cond-> (name k)
               *escape-?*
               (str/replace "?" "??"))]
-      (if (= \' (first n))
+      (if (str/starts-with? n "'")
         (let [ident   (subs n 1)
               ident-l (str/lower-case ident)]
           (binding [*quoted* (when-not (contains? #{"array"} ident-l) *quoted*)]
@@ -419,13 +419,16 @@
   ;; rather than name/namespace, we want to allow
   ;; for multiple / in the %fun.call case so that
   ;; qualified column names can be used:
-  (let [c (cond-> (str x) (keyword? x) (subs 1))]
-    (cond (= \% (first c))
+  (let [c (if (keyword? x)
+            #?(:clj (str (.sym ^clojure.lang.Keyword x)) ;; Omits leading colon
+               :default (subs (str x) 1))
+            (str x))]
+    (cond (str/starts-with? c "%")
           (let [[f & args] (str/split (subs c 1) #"\.")]
             [(str (format-fn-name f) "("
                   (join ", " (map #(format-entity (keyword %) opts)) args)
                   ")")])
-          (= \? (first c))
+          (str/starts-with? c "?")
           (let [k (keyword (subs c 1))]
             (cond *inline*
                   [(sqlize-value (param-value k))]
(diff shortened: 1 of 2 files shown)


src/honey/sql.cljc
```diff
@@ -161,13 +161,12 @@
...
 ;; suspicious entity names:
-(def ^:private suspicious #";")
-(defn- suspicious? [s] (boolean (re-find suspicious s)))
+(defn- suspicious? [s] (str/includes? s ";"))
 (defn- suspicious-entity-check [entity]
     (when-not *allow-suspicious-entities*
       (when (suspicious? entity)
         (throw (ex-info (str "suspicious character found in entity: " entity)
-                        {:disallowed suspicious})))))
+                        {:disallowed ";"})))))
```
seancorfield: Now you have the "magic" string `";"` in two places -- this should still use a `^:private` `def` for the string.

At some point, it may need to trap more than a single character so it may have to go back to either a regex or maybe a vector of substrings.

src/honey/sql/util.cljc
```diff
@@ -76,3 +76,14 @@
 
       :default
       (clojure.string/join separator (transduce xform conj [] coll)))))
+
+(defn into
+  "Drop-in replacement for `clojure.core/into` that doesn't perform
+  transient-persistent roundtrip. Given the usage patterns and data sizes that
+  HoneySQL deals with, not going through transients is more efficient."
```
seancorfield: I am slightly concerned that folks building large, complex SQL will be hitting thresholds where transients _are_ faster -- can you provide details on benchmarks for this, for both small and large SQL expressions (with 10s or maybe even 100s of parameters)?

src/honey/sql.cljc
```diff
@@ -658,11 +658,16 @@
...
         (format-expr x)))
 
-(defn- reduce-sql [xs]
-  (reduce (fn [[sql params] [sql' & params']]
-            [(conj sql sql') (if params' (into params params') params)])
-          [[] []]
-          xs))
+(defn- reduce-sql
+  ([xs] (reduce-sql identity xs))
+  ([xform xs]
```
seancorfield, line 664: Very nice! Thank you!

src/honey/sql.cljc
```diff
@@ -932,29 +934,33 @@
...
-                         (-> params (into params-j) (into u-params))])
-                      (let [[sql & params'] (when e (format-expr e))]
-                        [(cond-> sqls e (conj "ON" sql))
-                         (-> params
-                             (into params-j)
-                             (into params'))]))))
-                [[] []]
-                (partition-all 2 clauses))]
+        (transduce
+         (partition-all 2)
```
seancorfield, line 939: Nice!

src/honey/sql.cljc
alexander-yakushev: If that ever happens, it can go back to regular expressions (or something else). But for now, it makes sense to not pay for what's not needed.

src/honey/sql/util.cljc
alexander-yakushev: Well, you are right that it will be slower when there are 10s of parameters (the threshold is at about 10). It's not critical but should still be considered. Maybe, we can choose the better approach based on the number of elements in `from` – but I'm not a big fan of this approach.

Let me remove this commit from the PR for now and think about it more.

PR #554, opened 2024-11-22
Fixes #553 by adding :not-between

### LINKED ISSUE #553: Expand special syntax to include `NOT BETWEEN`
https://clojurians.slack.com/archives/C66EM8D5H/p1732295268360309

plooney81:
Extends special syntax to include `NOT BETWEEN`:

```clojure
(sql/format-expr [:not-between :id 1 100])
;;=> ["id NOT BETWEEN ? AND ?" 1 100]
```

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -1877,6 +1877,17 @@
         (into paramsx)
         (into params))))
 
+(defn- between-fn
+  "For both :between and :not-between"
+  [k [x a b]]
+  (let [[sql-x & params-x] (format-expr x {:nested true})
+        [sql-a & params-a] (format-expr a {:nested true})
+        [sql-b & params-b] (format-expr b {:nested true})]
+    (-> [(str sql-x " " (sql-kw k) " " sql-a " AND " sql-b)]
+        (into params-x)
+        (into params-a)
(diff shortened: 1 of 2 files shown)


src/honey/sql.cljc
```diff
@@ -1877,6 +1877,19 @@
         (into paramsx)
         (into params))))
 
+(defn- between-fn
+  "For both :between and :not-between"
+  [k [x a b]]
+  (let [[sql-x & params-x] (format-expr x {:nested true})
+        [sql-a & params-a] (format-expr a {:nested true})
+        [sql-b & params-b] (format-expr b {:nested true})
+        add-not?     (str/starts-with? (name k) "not")
```
seancorfield: Instead of the conditional stuff here, you can use `(sql-kw k)` which will turn `:between` into `BETWEEN` and `:not-between` into `NOT BETWEEN`.

seancorfield (changes requested):

src/honey/sql.cljc
plooney81: Will do!

seancorfield (approved):
Thank you! And thank you for adding documentation (which also runs as a test).

PR #559, opened 2024-12-28
fix broken links

kenhuang-io:


seancorfield:
Good catch. It got updated in two places previously, but I missed these two.

PR #560, opened 2024-12-29
More optimizations

alexander-yakushev:
Here are some more easy optimizations discovered while benchmarking Metabase's usage of HoneySQL.

1. `clojure.string/split` uses the regex machinery which is generally more wasteful than searching for a substring manually. HoneySQL so far only looks for `.` when splitting names. Besides, plenty of names don't have `.` in them and they still pay the regex tax. The new splitter function allocates nothing in such case.
2. Clojure's Keyword object already contains a corresponding Symbol, so we can return it in `kw->sym` (if there is no namespace). We could actually just call `(symbol kw)` for the same effect, but I think this behavior was introduced in Clojure 1.10, and HoneySQL claims to support 1.9.
3. I again tried to address the into's transient-persistent roundtrip. This time, I'm not making any speculative and potentially damaging changes. There are a lot of places in the code where `into` is used multiple times in succession, hence `into*` (name is up to be changed) both makes the code cleaner and prevents unnecessary transient roundtrips inbetween.

### DIFF
--- a/test/honey/util_test.cljc
+++ b/test/honey/util_test.cljc
@@ -43,3 +43,11 @@
   (is (= "1, 2, 3, 4"
          (sut/join ", " (remove nil?) [1 nil 2 nil 3 nil nil nil 4])))
   (is (= "" (sut/join ", " (remove nil?) [nil nil nil nil]))))
+
+(deftest split-by-separator-test
+  (is (= [""] (sut/split-by-separator "" ".")))
+  (is (= ["" ""] (sut/split-by-separator "." ".")))
+  (is (= ["hello"] (sut/split-by-separator "hello" ".")))
+  (is (= ["h" "e" "l" "l" "o"] (sut/split-by-separator "h.e.l.l.o" ".")))
+  (is (= ["" "h" "e" "" "" "l" "" "l" "o" ""]
+         (sut/split-by-separator ".h.e...l..l.o." "."))))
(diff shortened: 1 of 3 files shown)


alexander-yakushev:
Looks like `(.sym kw)` doesn't fly with Babashka. Perhaps, @borkdude may chime in if this is intentional?

borkdude:
@alexander-yakushev You can add this:

```
(:bb (symbol (name k)))
```

before the `:clj` branch

borkdude:
Btw since clojure 1.10 or so `(symbol kwd)` also just works I believe, so dropping clojure 1.9.0 support may also help

alexander-yakushev:
Yes, I mentioned it in the PR. I'm not sure if this optimization is a good reason to drop 1.9 if there are no others.

alexander-yakushev:
The custom `:bb` feature flag worked, thanks!

seancorfield:
Given 1.9 usage has dropped to just over 2.5%, I am considering dropping support for it. I'll ask in a few places to see how badly that would inconvenience folks...

test/honey/util_test.cljc
```diff
@@ -43,3 +43,20 @@
   (is (= "1, 2, 3, 4"
          (sut/join ", " (remove nil?) [1 nil 2 nil 3 nil nil nil 4])))
   (is (= "" (sut/join ", " (remove nil?) [nil nil nil nil]))))
+
+(deftest split-by-separator-test
```
seancorfield, line 47: I'd feel a bit more comfortable if these tests also compared against `(str/split <string> #"\.")` just to verify they're equivalent to that (slower) form that is being replaced in the source.

alexander-yakushev: There are differences between how `split-by-separator` and `str/split` behave in cases where there are leading, trailing, or consecutive separators. It is possible to make them behave the same at the cost of readability and implementation complexity, but I can't help wondering if it's worth it. After all, identifiers like `.foo..bar.` make no sense in HoneySQL, regardless of how we split the string.

seancorfield: Fair enough. OK, I'll merge this. Thank you!

PR #573, opened 2025-05-04
Performance improvements

alexander-yakushev:
Hi, here are a few improvements I discovered while working on Metabase.

The first commit addresses the following issue. Binding dynamic variables is comparatively quite expensive. `honey.sql/format` binds a lot of dynvars, but most of them usually don't receive meaningful values. A different approach is possible – prepare a bindings map that would contain only the vars with values that actually have changed, and pass that to `with-bindings` macro. Here are ad-hoc benchmark results obtained from Metabase tests:

```
Before
Time per call: 228.36 ms   Alloc per call: 559,342,831b
Time per call: 224.91 ms   Alloc per call: 559,342,813b
Time per call: 228.88 ms   Alloc per call: 559,342,813b

After
Time per call: 205.92 ms   Alloc per call: 452,848,231b
Time per call: 202.64 ms   Alloc per call: 452,848,232b
Time per call: 205.16 ms   Alloc per call: 452,848,229b
```

So, -20% allocations and -10% execution time.

The second commit ensures that `xs` passed to `format-values` is a vector because this function traverses `xs` multiple times using reduce, and the latter is inefficient for arbitrary seqs. The commit also contains a few minor changes that swap lazy seq processing for transducers.

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -1768,7 +1768,7 @@
   ([statement-map {:keys [aliased nested pretty]}]
    (binding [*options* (assoc *options* :dsl statement-map)]
      (let [[sqls params leftover]
-           (reduce (fn [[sql params leftover] k]
+           (reduce (fn [[sql params leftover :as result] k]
                      (if-some [xs (if-some [xs (k leftover)]
                                     xs
                                     (let [s (kw->sym k)]
@@ -1778,7 +1778,7 @@
                          [(conj sql sql')
                           (if params' (into params params') params)
                           (dissoc leftover k (kw->sym k))])
-                       [sql params leftover]))
+                       result))
                    [[] [] statement-map]
                    (:clause-order *options*))]
        (if (seq leftover)
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -1165,8 +1165,8 @@
             ;; use the keys from the first map if they match so that
             ;; users can rely on the key ordering if they want to,
             ;; e.g., see test that uses array-map for the first row
-            cols-n (into #{} (mapcat keys) (filter map? xs))
-            cols   (if (= (set cols-1) cols-n) cols-1 cols-n)]
+            cols-n (into #{} (comp (filter map?) (mapcat keys)) xs)
+            cols   (if (= (into #{} cols-1) cols-n) cols-1 cols-n)]
         [cols (when-not skip-cols-sql
                 (str "("
                      (join ", " (map #(format-entity % {:drop-ns true})) cols)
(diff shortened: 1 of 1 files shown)


seancorfield:
This seems to break `bb` and `cljs` compatibility.

alexander-yakushev:
I've made it pass on regular Clojure and Babashka, but it still doesn't work for Clojurescript because the latter lacks `with-bindings` macro. At this point, I see two possible ways forward:

1. Use `#?` to fallback to `bindings` for ClojureScript. Will complicate the code even further.
2. Since the dynvars in question are all part of the private API, it is possible to replace all of them with a single dynvar like `*options*` or something like that. Then, binding and reading a single dynvar would be much cheaper than a dozen.

I'd love to hear if you are comfortable with either approach.

seancorfield:
Option 2 sounds reasonable. As you say, all the dynvars that back the options are private, so no one should be depending on them. I believe there are a few places in the code that rebind some of the options, that will also need to change. Thank you in advance for doing this work.

alexander-yakushev:
Done. The changes seem to apply cleanly, didn't have to update any tests. 

I left `*dialect*` as a dedicated variable because its value is itself a map that has `:dialect` key, so putting it under `*options*` was confusing.

src/honey/sql.cljc
```diff
@@ -134,38 +134,41 @@
...
+   ;; can be added to format to turn this on:
+   :allow-suspicious-entities false
+   ;; the following metadata is ignored in formatted by default:
+   ;; :file, :line, :column, :end-line, and :end-column
+   ;; this dynamic var can be used to add more metadata to ignore:
+   :ignored-metadata []
+   ;; "linting" mode (:none, :basic, :strict):
+   :checking @default-checking
+   ;; the current DSL hash map being formatted (for clause-body / contains-clause?):
+   ;; caching data to detect expressions that cannot be cached:
```
seancorfield: Looks like this line got duplicated from L160 below and should be deleted.

src/honey/sql.cljc
```diff
@@ -419,29 +425,30 @@
...
 
 (defn ->numbered-param [k]
-  (let [n (count (swap! *numbered* conj k))]
+  (let [{:keys [numbered]} *options*
+        n (count (swap! numbered conj k))]
     [(str "$" n) (with-meta (constantly (dec n))
                    {::wrapper
-                    (fn [fk _] (param-value (get @*numbered* (fk))))})]))
-
-(def ^:private ^:dynamic *formatted-column* (atom false))
```
seancorfield, line 444: Interesting that this became unused at some point...

src/honey/sql.cljc
```diff
@@ -2283,50 +2292,38 @@
...
+                                (:quoted opts)
+                                (= :nrql (:dialect dialect))
+                                nil
+                                dialect?
+                                true
+                                :else
+                                @default-quoted)
+                  :quoted-always (:quoted-always opts @default-quoted-always)
+                  :quoted-snake (:quoted-snake opts @default-quoted-snake)
+                  :params (not-empty
```
seancorfield: Why the addition of `not-empty` here?

seancorfield (changes requested):

src/honey/sql.cljc
alexander-yakushev: It was relevant for the previous approach; can be removed now.

src/honey/sql.cljc
alexander-yakushev: Do you want to look into history to see if it should be brought back?

seancorfield: I looked and couldn't find where the additional uses had disappeared -- and no tests failed so I suspect whatever this was introduced to solve got reimplemented a different way and this var was just never removed.

alexander-yakushev:
Added another small commit that shaves unnecessary allocations further down.

seancorfield (approved):

PR #587, opened 2025-07-29
Remove erroneous `tap>` from abe10a6a

julienvincent:
Looks like a development tap snuck into the latest release. This removes it.


seancorfield:
We have `tap>` in our production code at work but I agree it shouldn't be left in a library -- good catch!

PR #588, opened 2025-08-13
Explain double colon :: syntax for cast

devurandom:
Add a hint to invoke `cast` instead of using the double colon `::` syntax.

I could not find an explanation how to render the `col::type` syntax and it took an excursion to the history of the Slack `#honeysql` channel to figure out what to do.  This documents my findings.

References: https://clojurians.slack.com/archives/C66EM8D5H/p1702579675520789?thread_ts=1702577803.253389&cid=C66EM8D5H


seancorfield:
Thanks for the prompt. It should probably be at the top of the `cast` section and I would use different wording, but your PR can act as a placeholder that.

PR #598, opened 2026-05-20
CASE docs: postgres also supports case-expr

holyjak:
IMO worth mentioning since Postgres is more wide-spread than BigQuery.


seancorfield:
Thank you!

PR #599, opened 2026-06-09
Add window frame support via :frame clause

yenda:
## What

Adds window **frame** support to HoneySQL — the `ROWS` / `RANGE` / `GROUPS … PRECEDING/FOLLOWING/CURRENT ROW/EXCLUDE` portion of a window specification — which was previously unsupported. It works inside both `:over` expressions and named `:window` definitions.

This clause is already in successful internal use in our project, where it backs real window-frame queries.

Window frames are ANSI-standard SQL (SQL:2003+) — this is **not** dialect-specific.

## Syntax

A new `:frame` clause (and matching `frame` helper) takes a positional vector in SQL token order:

```
:frame [mode (:between?) start (end) (exclusion?)]
```

- **mode**: `:rows`, `:range` or `:groups`
- **bounds**: `:unbounded-preceding`, `:current-row`, `:unbounded-following`, or an offset pair `[n :preceding]` / `[n :following]`
- **exclusion** (optional, trailing): `:exclude-current-row`, `:exclude-group`, `:exclude-ties`, `:exclude-no-others`

```clojure
;; data DSL
{:select [[[:over [[:sum :salary]
                   {:partition-by [:department]
                    :order-by     [:salary]
                    :frame [:rows :between :unbounded-preceding :current-row]}
                   :running-total]]]]
 :from :employee}
;; => SELECT SUM(salary) OVER (PARTITION BY department ORDER BY salary ASC
;;            ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS running_total FROM employee

;; offset bounds are expressions, so they parameterize (or inline):
{:frame [:range :between [5 :preceding] [10 :following]]}
;; => RANGE BETWEEN ? PRECEDING AND ? FOLLOWING   params: [5 10]

;; helper form composes by threading, like partition-by / order-by:
(over [[:sum :salary]
       (-> (partition-by :department)
           (order-by :salary)
           (frame :rows :between :unbounded-preceding :current-row))
       :running-total])
```

## Design notes

The syntax leans entirely on existing HoneySQL conventions rather than introducing new machinery:

- **`sql-kw`** already renders multi-word tokens: `:unbounded-preceding` → `UNBOUNDED PRECEDING`, `:current-row` → `CURRENT ROW`, `:exclude-no-others` → `EXCLUDE NO OTHERS` — the same mechanism behind `:select-distinct` / `:nulls-first`.
- **Offset bounds** `[n :preceding]` mirror `order-by`'s `[expr :dir]` pairs and flow through `format-expr`, so they support parameters, `:inline`, named params, and arbitrary expressions for free.
- The clause registers **between `:order-by` and `:limit`** in `default-clause-order`, so it renders in the correct position inside both `:over` (via the `format-dsl` window spec) and `:window` definitions with no changes to either formatter — the same global-registration approach `:partition-by` uses.
- The `:between` keyword is the **sole discriminator** between the single-start and two-bound forms (exact-keyword test, not arity), and exclusions are matched against a **closed set** so a typo errors loudly instead of leaking bad SQL. Malformed frames (bad mode, wrong bound count, unknown exclusion, trailing junk) throw `ex-info`.

## Tests

- `honey.sql-test/window-frame-tests` — format level: all three modes, single/`:between` bounds, offset params + `:inline` + named params, all four exclusions, the single-bound-plus-exclusion ambiguity case, frame inside `:over` and inside `:window`, and malformed-frame errors.
- `honey.sql.helpers-test/window-frame-helper-tests` — helper builds the clause vector and threads through `over`.
- `doc/clause-reference.md` gains a **frame** section; the examples are executed by the doc-block test suite.

Full suite green: JVM `clojure -M:test` (174 tests / 813 assertions), Babashka `bb test:bb` (173 / 794), `bb doc-test` (254 / 444), `bb eastwood` (0 warnings).

---

## Notes for review — two judgment calls

A couple of decisions here are genuine choices rather than the only option, so flagging them explicitly:

1. **Doc placement of the `frame` section.** I put `## frame` in the window cluster (right after `## window, partition-by (and over)`), not in strict `default-clause-order` position (which would be *after* `## order-by`). The reference doc already groups by concept — `## distinct, expr` sits in the same cluster, described as "related to the windowing clauses," despite `:distinct` being far earlier in clause-order — so keeping the window family together reads better. If you'd prefer strict clause-order, it's a one-block move below `## order-by`.

2. **Whether to validate at all.** `format-frame` throws `ex-info` on a bad mode, wrong bound count, or unknown exclusion. This follows the `:join-by` precedent (which validates a join type against an allowed set and throws unconditionally), since a malformed frame vector can't render meaningful SQL. But it's slightly stricter than pure pass-through; if you'd rather emit whatever and let the database reject it, the validation is easy to drop. Happy to go either way.

Verified across the JVM, Babashka **and ClojureScript** suites (the cljs run confirms the `thrown-with-msg? ExceptionInfo` assertions and the `.cljc` code compile and pass under cljs), plus `bb eastwood` (0 warnings) and `bb doc-test`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changes
 
 * 2.7.next in progress
+  * Add support for window frames via the `:frame` clause (and `frame` helper) inside `:over` expressions and `:window` definitions, e.g. `{:frame [:rows :between :unbounded-preceding :current-row]}` produces `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`. Supports `:rows`/`:range`/`:groups` modes, single or `:between` bounds (`:unbounded-preceding`, `:current-row`, `:unbounded-following`, or `[n :preceding]`/`[n :following]` offset pairs), and frame exclusions.
   * Note that PostgreSQL also supports `CASE` via PR [#598](https://github.com/seancorfield/honeysql/pull/598) from [@holyjak](https://github.com/holyjak).
   * Upgrade from EPL-1.0 to EPL-2.0 for broader compatibility. Explicitly allow for relicensing of HoneySQL code under Apache-2.0.
   * Update dev/test deps.
--- a/doc/clause-reference.md
+++ b/doc/clause-reference.md
@@ -20,7 +20,8 @@ The examples herein assume:
 (refer-clojure :exclude '[partition-by])
 (require '[honey.sql :as sql]
          '[honey.sql.helpers :as h :refer [select from join-by left-join join
-                                           where order-by over partition-by window]])
+                                           where order-by over partition-by window
+                                           frame]])
 ```
 
 Every DDL and SQL clause has a corresponding helper function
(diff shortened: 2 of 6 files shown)


seancorfield:
Thanks, @yenda (and Claude!). I updated the CHANGELOG to include links to your PR and your username page (and acknowledge co-authorship with Claude).

PR #600, opened 2026-06-26
[perf] Optimize suspicious?

alexander-yakushev:
Reclaim some of the efficiency that was lost as `suspicious?` became more complex to address the CVE in 2.7.1392.

The main benefit is avoiding turning characters into strings (which is slow and allocates).

Quick benchmark from #545 shows this:

```
Before:
Time per call: 48.60 us   Alloc per call: 72,360b
Time per call: 48.63 us   Alloc per call: 72,360b
Time per call: 48.72 us   Alloc per call: 72,360b

After:
Time per call: 44.46 us   Alloc per call: 66,136b
Time per call: 44.31 us   Alloc per call: 66,136b
Time per call: 44.32 us   Alloc per call: 66,136b
```

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -170,8 +170,13 @@
 (def ^:no-doc ^:dynamic *nest-infix* true)
 
 ;; suspicious entity names:
-(def ^:private suspicious ";,")
-(defn- suspicious? [s] (some #(str/includes? s (str %)) suspicious))
+(def ^:private suspicious (vec ";,"))
+(defn- suspicious? [s]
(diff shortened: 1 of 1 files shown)


seancorfield:
Thanks! I was hoping you'd take a look and offer a performance improvement. I did wonder if a single regex via `re-find` would be faster than `some`, given there are only two characters?

alexander-yakushev:
Thanks for merging! Regex is still slower, I've checked. But if the check becomes yet more complicated, reverting to a regexp may become a better solution.

PR #601, opened 2026-06-30
[perf] Further optimize suspicious?

bsless:
The existing `suspicious?` still added two performance overheads: iteration and allocation.

Both can be pushed to eval step instead of runtime as long as the suspicious check remains one of simple char membership

Before:
```
44.2 ns  3σ [15.8 72.6]  min 40.4
69.3 ns  3σ [27.8 111]  min 38.5
44.5 ns  3σ [20.0 69.1]  min 37.8
```
After:
```
7.61 ns  3σ [6.96 8.26]  min 6.81
9.68 ns  3σ [8.40 11.0]  min 8.80
8.79 ns  3σ [3.76 13.8]  min 6.63
```

I've been following all perf related PRs and thought I might chime in :)

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -31,7 +31,7 @@
   (:require [clojure.string :as str]
             #?(:clj [clojure.template])
             [honey.sql.protocols :as p]
-            [honey.sql.util :refer [str join split-by-separator into*]]))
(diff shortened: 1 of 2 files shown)


seancorfield:
Oh, that's nice! Thanks, @bsless

PR #602, opened 2026-06-30
[perf] Rewrite clj alphanumeric check to state machine

bsless:
The current alphanumeric regex is simple enough to translate to a state
machine.

The hand rolled state machine implementation is about 10x faster, see
added RCF added at the test file.

Added property based tests to guard against regression in the rewrite.

Tests added only for clj, requires test-chuck which bb can't run, so
added another alias which is appended for non bb non cljs tests

### DIFF
--- a/bb.edn
+++ b/bb.edn
@@ -46,7 +46,7 @@
            :extra-deps  {io.github.cognitect-labs/test-runner
                          {:git/tag "v0.5.1" :git/sha "dfb30dd"}}
            :task (exec 'cognitect.test-runner.api/test)
-           :exec-args {:patterns ["^(?!honey.cache).*-test$"]}}
+           :exec-args {:patterns ["^(?!(honey.cache|honey.sql-alphanumeric)).*-test$"]}}
 
   test:all {:doc "Run all tests for different Clojure versions and ClojureScript"
             :task (binding [*command-line-args* (conj *command-line-args* "all")]
--- a/deps.edn
+++ b/deps.edn
@@ -20,7 +20,8 @@
   {:extra-paths ["test"]
    :extra-deps  {io.github.cognitect-labs/test-runner
                  {:git/tag "v0.5.1" :git/sha "dfb30dd"}
-                 org.clojure/core.cache {:mvn/version "1.2.263"}}
+                 org.clojure/core.cache {:mvn/version "1.2.263"}
+                 com.gfredericks/test.chuck {:mvn/version "0.2.15"}}
    :exec-fn     cognitect.test-runner.api/test}
   :runner
   {:main-opts   ["-m" "cognitect.test-runner"]}
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -310,6 +310,79 @@
 
 (def ^:private ^:dynamic *drop-ns* false)
 
+#?(:clj
+   (defn alphanumeric?
+     "Translation of the prior alphanumeric regex to state machine.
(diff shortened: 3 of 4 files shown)


bsless:
Checking what went wrong with the build

bsless:
@seancorfield the property based tests are failing due to weird characters that match `$` such as `\205`, `\r` and `\n`
Do you think the regex requires refinement, using a different anchor such as `\z` or `\Z`?

seancorfield:
Looks like you solved it by stripping "weird" characters in the PBT. Could you use a character construction from the int code for `\205` so we don't have a "funky" char in the string that might get borked by various editors that future maintainers use?

bsless:
Done. I'm still afraid of other funky characters, but at least it's readable now.

bsless:
Another thing I dislike about this PR is the test won't load without adding an alias.
One way to make the development experience more pleasant would be pulling the PBT to a separate ns, wdyt?

seancorfield:
Since it is a Clojure-only test, moving it to `sql_alphanumeric_test.clj` would seem like a good option, yes.

bsless:
How will I make babashka skip it? Reader conditionals aren't allowed in clj files. I can name it cljc to regain them but that seems excessive.

bsless:
Figured out the least disgusting solution

test/honey/sql_test.cljc
```diff
@@ -4,6 +4,13 @@
   (:refer-clojure :exclude [format])
   (:require [clojure.string :as str]
             [clojure.test :refer [deftest is testing]]
+            #?@(:bb []
```
seancorfield: Seems like this whole block is no longer needed?

bb.edn
```diff
@@ -40,13 +40,13 @@
...
                         ":test:"
-                        (if cljs? "cljs" "runner")))))}
+                        (if cljs? "cljs" "spec:runner")))))}
 
   test:bb {:extra-paths ["test"]
            :extra-deps  {io.github.cognitect-labs/test-runner
                          {:git/tag "v0.5.1" :git/sha "dfb30dd"}}
            :task (exec 'cognitect.test-runner.api/test)
-           :exec-args {:patterns ["^(?!honey.cache).*-test$"]}}
+           :exec-args {:patterns ["^(?!(honey.cache|honey.sql-alphanumeric-test)).*-test$"]}}
```
seancorfield: `honey.sql-alphanumeric` -- doesn't need `-test` here (like `honey.cache` doesn't have).

deps.edn
```diff
@@ -20,8 +20,11 @@
   {:extra-paths ["test"]
    :extra-deps  {io.github.cognitect-labs/test-runner
                  {:git/tag "v0.5.1" :git/sha "dfb30dd"}
-                 org.clojure/core.cache {:mvn/version "1.2.263"}}
+                 org.clojure/core.cache {:mvn/version "1.2.263"}
+                 com.gfredericks/test.chuck {:mvn/version "0.2.15"}}
```
seancorfield, line 24: Not needed here, given it is in `:spec`?

bsless: Since I split it off to an ignored namespace, I'm considering removing the `:spec` alias, wdyt?

seancorfield: As long as it still works with `bb` and `cljs`, sure. It's simpler than using an alias.

bsless: Works and looks better.

seancorfield:
Much appreciated!

PR #603, opened 2026-07-12
[perf] Check for suspicious entries first, then if they are allowed

alexander-yakushev:
Since `:allow-suspicious-entities` defaults to false, and will rarely be enabled, almost every caller pays the extra cost of dereferencing a dynamic variable. And because most entities are not suspicious, then this dereference will not be needed at all most of the time.

In my syntethic benchmark, I'm getting a 3% improvement on a flamegraph. On the benchmark results itself, the improvement is harder to spot (variance is high), but I think it is still worth doing.


seancorfield:
Good catch! Thanks.

PR #605, opened 2026-07-27
pin GHA deps by commit SHA

Totara-thib:
Hi Sean, small drive-by contribution: this pins the four actions used across the three workflows to their commit shas, versions kept as comments.

The one that motivated it is `DeLaGuardo/setup-clojure@master`: a branch ref, so every run executes whatever master points to at that moment, including in `test-and-release.yml` which runs with your Clojars deploy secrets. A compromised or hijacked upstream re-points the ref and your next release build runs their code. Same pattern as the tj-actions/changed-files incident (CVE-2025-30066). The pin is the current 13.6.1 release commit, which is exactly where master sits today, so nothing changes behavior.

The `actions/*` ones are lower risk but pinned too while at it: checkout `3d3c42e` is v7.0.1, setup-java `03ad4de` is v5.6.0, cache `55cc834` is v6.1.0. All shas resolved from the upstream repos and cross checked against their tags. No logic changes, only the `uses:` lines.

Found with the Plumber CLI (https://github.com/getplumber/plumber), verified on develop. I can also open a PR which will adds it to CI so this does not quietly drift back, that one is a bonus, this PR stands on its own.

### DIFF
--- a/.github/workflows/test-and-release.yml
+++ b/.github/workflows/test-and-release.yml
@@ -9,20 +9,20 @@ jobs:
   build-and-release:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 0
-      - uses: actions/setup-java@v5
+      - uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5.6.0
         with:
           distribution: 'temurin'
           java-version: '17'
       - name: Setup Clojure and Babashka
-        uses: DeLaGuardo/setup-clojure@master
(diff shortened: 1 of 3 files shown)


seancorfield:
Thanks. I've pinned `setup-clojure` to a full SHA. I'm talking with some folks in the Clojure community who are looking at supply chain integrity issues and have shown them this PR and your tool.

Since the other `actions` are all GitHub's own (and therefore Microsoft's own), I'm leaving those as-is for now.

A disadvantage to switching to SHAs is that tools for checking outdated dependencies then switch to looking at commits and will "recommend" the latest SHA on the main branch, regardless of whether it corresponds to a tagged release or not, so it creates a maintenance burden (so it's a tradeoff).

seancorfield:
I'm also working my way through other projects where I rely on `setup-clojure` and pinning those to full SHAs as well, so thank you for the prompt on that.

PR #609, opened 2026-08-22
[perf] Promote :inline to a separate dynvar

alexander-yakushev:
Many places in the code rebind `*options*` only to enable :inline. This pays the cost with `(assoc *options* :inline true)` and with creating a hash-map inside `*binidng*`. This PR creates a separate `*inline*` dynvar, and uses a more efficient way to bind it – we pre-construct the necessary hash-map and use `push-thread-bindings` to set it.

I've benchmarked this and the next two PRs together on the #545 benchmark.

```
Before:
Time per call: 59.12 us   Alloc per call: 55,920b
Time per call: 59.86 us   Alloc per call: 55,920b
Time per call: 60.81 us   Alloc per call: 55,920b

After:
Time per call: 47.18 us   Alloc per call: 47,688b
Time per call: 47.99 us   Alloc per call: 47,688b
Time per call: 47.41 us   Alloc per call: 47,688b
```

~20% time and ~15% alloc improvement.

### DIFF
--- a/src/honey/sql.cljc
+++ b/src/honey/sql.cljc
@@ -138,6 +138,7 @@
 (def ^:private default-standard-conforming-strings (atom true))
 
 (def ^:private ^:dynamic *dialect* nil)
+(def ^:private ^:dynamic *inline* @default-inline)
 (def ^:private ^:dynamic *options*
   {;; nil would be a better default but that makes testing individual
    ;; functions harder than necessary:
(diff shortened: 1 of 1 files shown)


src/honey/sql.cljc
```diff
@@ -197,6 +197,15 @@
    (defn- suspicious? [s]
      (some (fn [ch] (str/includes? s (str ch))) suspicious)))
 
+(def ^:private inline-true-map {#'*inline* true})
+
+(defmacro with-inline
```
seancorfield: Please add `^:no-doc` metadata so this doesn't appear in cljdoc.org output.

PR #617, opened 2026-09-21
Simplify let-go CI setup

ingydotnet:
Install let-go and lgx directly through the in-1 bootstrap.

Remove the separate in-1 installation and update step.


seancorfield:
No, I had it like that originally and did not like it. The Broken pipe error happened in that version too.

